PR #46, opened 2023-06-28
Added specs for IG, IMP and IMPCON

jawache:
Attempted to merge the specs for the Impact Measurement Plugin and the Ontology Project into one.

From this point forward we will use markdown in the specs folder to define the spec.

### DIFF
--- a/specs/impact-graph.md
+++ b/specs/impact-graph.md
@@ -20,7 +20,7 @@ Concrete example of how we might model the carbon impact of a web server using a
 My Web Server Impact Graph
 ```
 
-We first create an **Impact Graph** that describes the ontology we want to measure. This might be an application like this case which is a Web Server, but it also might be a campaign or even an individual user journey. It could even be used to measure the embodied carbon of physical devices. What we are proposing here isn't limited to just software.
+We first create an **Impact Graph** that describes the software boundary we want to measure. The software boundary is represented as an ontology. This might be an application like this case which is a Web Server, but it also might be a campaign or even an individual user journey. It could even be used to measure the embodied carbon of physical devices. What we are proposing here isn't limited to just software.
 
 ```
 My Web Server Impact Graph
--- a/specs/impact-graph.md
+++ b/specs/impact-graph.md
@@ -1,0 +1,223 @@
+# Impact Graph: A Software Impact Model
+
+If you can’t measure, you can’t improve. Software has many negative environmental **impacts** which we need to optimize, carbon, water, energy to name just a few.
+
+Unfortunately, measuring software impact metrics like carbon, water, energy is complex and very nuanced. 
+
+Modern applications are composed of many smaller pieces of software (components) running on many different environments, for example, private cloud, public cloud, bare-metal, virtualized, containerized, mobile, laptops, desktops, embedded, IoT. Many components that make up a typical software application are not run on resources you own or control, which makes including the impact of managed services in your measurement especially hard. 
+
+- How we model the environmental impacts of a complex application is through the creation of an **Impact Graph** which represents and models the thing we want to measure. 
+- Each **Impact Graph Node** in the graph represents one atomic thing we want to measure, it could be a software component, or part of the user joruney. 
+- To measure the environmental impact of one of the nodes in an **Impact Graph** we use a **Impant Mesurement Model**. 
+- There are many **Impact Measurement Models** out there and they all have different interfaces, some are APIs some are libraries, some are just CSV files. 
+- Since every **Impact Measurement Model** is bespoke to make them all interoperable we write **Impact Measurement Plugins** which adhere to a common **Impact Measurement Interface** specification.
+
+### Example
+
+Concrete example of how we might model the carbon impact of a web server using an **Impact Graph**.
+
+```
+My Web Server Impact Graph
+```
+
+We first create an **Impact Graph** that describes the ontology we want to measure. This might be an application like this case which is a Web Server, but it also might be a campaign or even an individual user journey. It could even be used to measure the embodied carbon of physical devices. What we are proposing here isn't limited to just software.
+
+```
+My Web Server Impact Graph
+├── Backend Server
+├── Load Balencer
+└── Cache Servers
+```
+
+We second break out the ontology into the nodes which we want to measure. The important thing is that we are breaking things out not as concepts, but as seperate and distint measurable components of the whole. 
+
+An **Impact Graph Node (IGN)** is something we want to measure distinct from every other node in the Impact Graph. That's it, we can clump all the servers together into one IGN regardless of what's running on them IF we want to measure them all together. We can break it up into one IGN per-process IF we want to measure per-process. The IGN describes an atomic thing we want to measure, that's it.
+
+What this means is that although a component might be "Backend Servers" since we'd use different approaches to measure "Intel Backend Servers" and "AMD Backend Servers" we might actually break it out into two IGNs, "AMD Backend Servers" and "Intel Backend Servers". This sounds complex but since our goal is to measure, measuring dictates how we strucutre the graph.
+
+```
+My Web Server Impact Graph
+├── Backend Server <- class BoavistaVM extends IMI -> Boavista IMM
+├── Load Balencer <- class BoavistaVM extends IMI -> Boavista IMM
+└── Cache Servers <- class IntelVM extends IMI -> Intel IMM
+```
+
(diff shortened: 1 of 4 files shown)


jawache:
Discussed in Project call, people to review the content and we'll revisit next meeting.

navveenb:
@rish05, can you also please review this pull request

specs/impact-graph.md
```diff
@@ -0,0 +1,223 @@
...
+
+
+```
+My Web Server Impact Graph
+├── Backend Server <- class CCFVM extends IMI -> CCF IMM
+├── Load Balencer <- class CCFVM extends IMI -> CCF IMM
+└── Cache Servers <- class IntelVM extends IMI -> Intel IMM.
+```
+
+If later on we wanted to use another IMM, perhaps the Cloud Carbon Footprint IMM, we would
```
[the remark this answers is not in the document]
srini1978: Made then changes

specs/impact-model-plugin.md
```diff
@@ -0,0 +1,335 @@
...
+
+### TelemetryStream
+
+This data type returns a stream of Telemetry objects, used when we want to calculate impact metrics in real-time.
+
+### ImpactModel
+
+This datatype holds the instance for an IMO.
+
+### ImpactMetric
```
srini1978, line 61: We need to provide customer perspective of what an Impact metric is . For the end user (here in this case the software developer who is trying to measure the "impact" of his software), there should be a way for him/her to explicitly call out the impact metric they are trying to measure . Impact metric could be carbon, water, energy etc . We need to find a way to break this impact metric into some classes that can be leveraged directly by the customer on installing the Impact Measurement Plugin . This way, a person who is measuring carbon will be using classes that talk the "carbon" language (names of classes, methods etc) and someone who is measuring "water" will be using that appropriate modules.

As soon as the IMP is installed there should be a way to specify if the user is intending to measuring carbon, water, energy etc.  This should be given as input to the plugin along with the telemetry. 

For example 
pip install CloudCarbonFootprintplugin carbon

This will use the carbon related impact metric classes. The carbon related impact metric classes will probably derive from ImpactMetric class or interface.

srini1978 (approved):
Have incorporated 2 of the 3 review comments. The third comment is something around providing a end user angle to the impact metric and make it more customizable to someone who is measuring carbon vs water.

Will try to think of something and come up with ideas before the weekly carbon call

specs/impact-graph.md
```diff
@@ -0,0 +1,223 @@
...
+
+### Example
+
+Concrete example of how we might model the carbon impact of a web server using an **Impact Graph**.
+
+```
+My Web Server Impact Graph
+```
+
+We first create an **Impact Graph** that describes the ontology we want to measure. This might be an application like this case which is a Web Server, but it also might be a campaign or even an individual user journey. It could even be used to measure the embodied carbon of physical devices. What we are proposing here isn't limited to just software.
```
navveenb, line 23: We first create an **Impact Graph** that describes the software boundary we want to measure. The software boundary is represented as an ontology.

specs/impact-graph.md
```diff
@@ -0,0 +1,223 @@
...
+├── Backend Server
+├── Load Balencer
+└── Cache Servers
+```
+
+We second break out the ontology into the nodes which we want to measure. The important thing is that we are breaking things out not as concepts, but as seperate and distint measurable components of the whole. 
+
+An **Impact Graph Node (IGN)** is something we want to measure distinct from every other node in the Impact Graph. That's it, we can clump all the servers together into one IGN regardless of what's running on them IF we want to measure them all together. We can break it up into one IGN per-process IF we want to measure per-process. The IGN describes an atomic thing we want to measure, that's it.
+
+What this means is that although a component might be "Backend Servers" since we'd use different approaches to measure "Intel Backend Servers" and "AMD Backend Servers" we might actually break it out into two IGNs, "AMD Backend Servers" and "Intel Backend Servers". This sounds complex but since our goal is to measure, measuring dictates how we strucutre the graph.
```
navveenb, line 36: We can add Cloud to the above details - measure "Intel Backend Servers" and "AMD Backend Servers"  or "Cloud Backend VMs"

specs/impact-graph.md
```diff
@@ -0,0 +1,223 @@
...
+
+```
+My Web Server Impact Graph
+├── Backend Server <- class CCFVM extends IMI -> CCF IMM
+├── Load Balencer <- class CCFVM extends IMI -> CCF IMM
+└── Cache Servers <- class IntelVM extends IMI -> Intel IMM.
+```
+
+
+All IMPs expose the same class Interface (IMI) so can be swapped in and out easily.
```
navveenb, line 61: All IMPs extend from the same base class Interface (IMI), so can be swapped in and out easily. New IMPs can be added by Organizations/Users by extending the base class Interface.

specs/impact-model-plugin.md
```diff
@@ -0,0 +1,335 @@
...
+
+Telemetry can be aggregate or granular, for example it might contain an average of cpu utilization across a range of machines.
+
+`{ “date-time”: xxxx, “duration”: xxx, “cpu-util”: 0.5 }`
+
+Or it might contain more granular data for every machine like so:
+
+`{ “date-time”: xxxx, “duration”: xxx, machines: { “gsf-001”: “cpu-util”: 0.5, “gsf-002”:  “cpu-util”: 0.62 } }`
+
+Or it might contain even more granular data to the process level:
```
navveenb, line 47: We probably need extend the example with machineType, like a Cloud VM type like D2, C2 machines, or On-premises with actual machine spec.

specs/impact-model-plugin.md
jawache: I suspect where we will end up is the impact will be determined by the plugin. So you'd install the plugin (IMP) for the specific method of measurement you want.

so com.thoughworks.ccf.carbon.sci
so com.thoughworks.ccf.carbon.ghg
so com.thoughworks.ccf.water

I suspect the whole methodology of the plugin, IMP, IMM will be fundamentally different for every approach. 

Although I don't think there is anything we are saying that would stop people from just writing a generic IMP that does many different things. I personally think they should be narrowly focused on a specific area.

If a plugin does do multiple impacts it could be configured with the params that are passed in when we instantiate this plugin (e.g. mode: ghg). But I do think we should somehow have some guidance where each plugin does one unique impact metric. Plugins that can be configured to return multiple impacts are going to be a huge headache down the line when people forgot to configure with the right impact mode and complain when one node returns water and another node returns sci.

Or actually maybe you need to declare at the top of your Impact Graph the type of Impact Metric you are measuring and then we can verify/validate that the plugins you are trying to us actually return that impact metric.

specs/impact-graph.md
jawache:
```suggestion
We first create an **Impact Graph** that describes the software boundary we want to measure. The software boundary is represented as an ontology. This might be an application like this case which is a Web Server, but it also might be a campaign or even an individual user journey. It could even be used to measure the embodied carbon of physical devices. What we are proposing here isn't limited to just software.
```

specs/impact-model-plugin.md
jawache: Agreed, I think as we start developing Telemetry is going to have much more details, suspect it's going to have it's own spec. I think this will form a summary doc and with that in mind I'd suggest keeping this initial example simple.

specs/impact-graph.md
jawache:
```suggestion
All IMPs extend from the same base class Interface (IMI), so can be swapped in and out easily. New IMPs can be added by Organizations/Users by extending the base class Interface.
```

jawache (commented):
Added some suggested edits based on feedback, we can review on the next call.

specs/impact-graph.md
jawache: Oh my intention was Cloud. I think this is an interesting point would like to delve in with you and make sure we are on the same page.

We've discussed in previous ontology discussions and examples about a node being a single "software component" or at least that's how I felt each node in an ontology represented.

But as I've been thinking about this more and as we've been exploring this more at Intel a node in an ontology is not really a software component for us, it's a _thing_ we want to measure _separate_ from everything else we are measuring. 

So at Intel, we might want to work with a client and break a software component into the processes that run on AMD vs the processes that tun on Intel. So we can see how much energy the same process consumes when running on AMD chips vs Intel chips.

It's subtle but I think that's it's an important characteristic of the ontology, the lowest level of granularity isn't the process that's run, it's just a clustering of things we want to measure and that might not even relate to a process. Does that make sense?

jawache:
Team: Agreed to merge as is, we will add to it as we go along.

PR #51, opened 2023-07-02
Proposed spec for impact yaml

jawache:
A proposed yaml based file format to define a formal way for writing an impact graph/ontology. 
Useful as a format to capture example ontologies which we can use to drive the development of this project.

As I start breaking down this project into tasks we really need a way to formally describe a "use case" e.g. an application, components, imps, params and telemetry which the end application will have to calculate. With impl we can work in parallel and create impls for a variety of use cases (maybe translate some of the existing hand written ones?) whilst development happens on the framework itself.

### DIFF
--- a/specs/impact-yaml.md
+++ b/specs/impact-yaml.md
@@ -1,0 +1,169 @@
+# Impact Yaml (IMPL)
+
+- Impact Yaml (or IMPL) is a file format based off [yaml](https://circleci.com/blog/what-is-yaml-a-beginner-s-guide/) that models an ontology you want to meaure.
+- IMPL is the file format Impact Graphs can be saved to (and loaded from).
+- IMPL being YAML means that it's more human readable and can be used as a **formal method of writing use cases**, such as SCI use cases.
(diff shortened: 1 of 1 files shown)


gnanakeethan:
This method of input is really good. 

We should be able to define a schema that is extensible across different IMPs as each and every IMP might require different inputs.

jawache:
Agreed @gnanakeethan I do think every IMP might have different inputs, it's why I think we have to stay flexible and make the inputs a dictionary (I have no idea what a water model might require as inputs!)

Do you know is there a way of defining a schema for a yaml, I know there is a way for XML. So a parser can statically validate an IMPL file to make sure it's formed correctly before parsing iinto code?

jawache:
Naveen: Focus on SCI first before GHG.
Gadhu: Models can be calibrated using this method. Calibrate GHG vs SCI models. 
Gadhu: GHG models might only be used for yearly reporting, which means yearly telemetry. SCI might be preferable for shorter timeframes.

jawache:
Project team agreed to merge the spec

Issue #49, opened 2023-07-01
IMP POC for Boavizta


jawache:
**What**: Create an initial PoC for an IMP which interfaces with the Boavizta API.
**Why**: So we can validate the specification and learn from implementation.
**SoW**:
- Create a concrete class that implements our IMI interface to the Boavizta API
- It should have some unit tests that call the class with sample data and prints out the response (note in future we will validate code like that by using the impconsole)

@gnanakeethan as discussed let's approach this task like this:

- [ ] #55 (@jawache to gather)
- [x] Create a `BoaviztaIMP `class which implements the class interface as defined in https://github.com/Green-Software-Foundation/carbon-ql/blob/dev/specs/impact-model-plugin.md
**NOTE**: As discussed let's not strongly type the inputs, keep them as dictionaries/objects. I know the spec has type names but let's assume this as just alias's for object. I'd like us to delay strongly typing things for as long as possible (until we have to). Communicate if the params are wrong type/shape using exceptions.
- [x] Create some code which implements `BoaviztaIMP`  and inputs telemetry etc... and prints output.
- [x] Use JSII as per #43 to create the same code in multiple languages and check the actual generated cod  in as well under different folders.
- [x] Create a pull request with all the files and submit that for review the next meeting. 
**NOTE** the pull request should let us view, compare and comment on the source TS code and the generated code. We'll discuss next call and encourage others to add their comments to the PR.

jawache:
@gnanakeethan as discussed let's approach this task like this:

- [ ] Gather real telemetry examples from the SCI case studies
- [ ] Create a `BoaviztaIMP `class which implements the class interface as defined in https://github.com/Green-Software-Foundation/carbon-ql/blob/dev/specs/impact-model-plugin.md
**NOTE**: As discussed let's not strongly type the inputs, keep them as dictionaries/objects. I know the spec has type names but let's assume this as just alias's for object. I'd like us to delay strongly typing things for as long as possible (until we have to). Communicate if the params are wrong type/shape using exceptions.
- [ ] Create some code which implements `BoaviztaIMP`  and inputs telemetry etc... and prints output.
- [ ] Use JSII as per #43 to create the same code in multiple languages and check the actual generated cod  in as well under different folders.
- [ ] Create a pull request with all the files and submit that for review the next meeting. 
**NOTE** the pull request should let us view, compare and comment on the source TS code and the generated code. We'll discuss next call and encourage others to add their comments to the PR.

[...]

PR #56, opened 2023-07-11
JSII Implementation Preview

gnanakeethan:


gnanakeethan:
Unit Testing is not done. 

Potential issue can exist in Promise resolution for Axios Request.

Generated Golang code is much readable than generated JS / Python code.

gnanakeethan:
There is no typing available for Errors. https://github.com/aws/jsii/issues/369

jawache:
Thanks @gnanakeethan. I've put a few initial comments in please take a look.

Also it's quite hard to dissect and look at the generated code due to the sheer quantity of the generated code in the TypeScript files. I know in the spec we have typed every input but lets stick to simple object (or whatever the correct built-in type for a dictionary is). I suspect that will return the amount of generated code and allow us to evaluate easier.

Also can we have Java, .Net and Python as the generated languages. I think those are going to be the most useful and well understood by all to evaluate. No need for Go at this stage.

jawache (changes requested):

lib/interfaces/index.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+
+export interface IImpactModelInterface {
+    plugin: IImpactPluginInterface
+
+    create(name: string, staticParams: IStaticParams | undefined): IImpactModelInterface
+
+    authenticate(authParams: AuthParams): void
+
+    snapshot(data: ITelemetryInterface): Promise<IImpactMetricInterface>
```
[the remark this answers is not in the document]
gnanakeethan: changed to object type

lib/interfaces/index.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+
+export interface IImpactMetricStream {
+}
+
+export interface AuthParams {
+}
+
+export interface IImpactModelInterface {
+    plugin: IImpactPluginInterface
```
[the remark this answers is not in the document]
gnanakeethan: it is not possible to implement use "fetchData" function from Plugin if we do not define a ImpactPluginInterface.

lib/boavizta.ts
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    stream(data: ITelemetryStream): IImpactMetricStream {
+        console.log(data);
+        return {};
+    }
+}
+
+
+export class BoaviztaCloudPlugin implements IImpactPluginInterface {
```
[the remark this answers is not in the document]
gnanakeethan: it is not possible to implement use "fetchData" function from Plugin if we do not define a ImpactPluginInterface.

lib/boavizta.ts
```diff
@@ -0,0 +1,72 @@
...
+    batch(data: ITelemetryInterface[]): Promise<IImpactMetricInterface[]> {
+        return Promise.all(data.map((d) => this.plugin.fetchData(d)));
+    }
+
+    authenticate(authParams: AuthParams): void {
+        this.authParams = authParams;
+        console.log(this.authParams)
+    }
+
+    create(name: string, staticParams: IStaticParams | undefined = undefined): IImpactModelInterface {
```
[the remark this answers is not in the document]
gnanakeethan: changed to object type

srini1978:
@gnanakeethan  With respect to your question on dotnet , the 2 packages that are deployed with *.pkg have to be pushed to Nuget repository to be able to see the dotnet files i believe

https://aws.github.io/jsii/user-guides/lib-author/configuration/targets/dotnet/

lib/boavizta.ts
```diff
@@ -0,0 +1,135 @@
...
+}
+
+export class BoaviztaCloudImpactModel implements IImpactModelInterface {
+    public provider: string | undefined;
+    public name: string | undefined;
+    protected authCredentials: IBoaviztaAuthCredentials | undefined;
+
+    create(name: string, staticParams: object | undefined): object {
+        const staticParamCast = staticParams as IBoaviztaStaticParams;
+        if (staticParamCast?.hasOwnProperty('provider')) {
```
jawache, line 30: Is there a good reason you are using hasOnwPropery instead of `in`, 

if ("provider" in statisParams) is much more succint.

lib/boavizta.ts
```diff
@@ -0,0 +1,135 @@
...
+export interface IBoaviztaAuthCredentials {
+}
+
+export class BoaviztaCloudImpactModel implements IImpactModelInterface {
+    public provider: string | undefined;
+    public name: string | undefined;
+    protected authCredentials: IBoaviztaAuthCredentials | undefined;
+
+    create(name: string, staticParams: object | undefined): object {
+        const staticParamCast = staticParams as IBoaviztaStaticParams;
```
jawache, line 29: We've discussed and it's documentation in the issues, just to use `object` for now to keep things simple.

lib/boavizta.ts
```diff
@@ -0,0 +1,135 @@
+import {IImpactModelInterface, IStaticParams} from "./interfaces/index";
+import axios from "axios";
+
+export {
+    IImpactModelInterface,
+    IStaticParams
+} from "./interfaces/index";
+
+
+export interface IBoaviztaStaticParams extends IStaticParams {
```
jawache, line 10: Remove

lib/boavizta.ts
```diff
@@ -0,0 +1,135 @@
...
+    provider?: string;
+    componentType?: string;
+}
+
+export interface IBoaviztaCloudTelemetry {
```
jawache, line 15: Remove

lib/boavizta.ts
```diff
@@ -0,0 +1,135 @@
...
+
+}
+
+
+export interface IBoaviztaAuthCredentials {
```
jawache, line 20: Remove

lib/boavizta.ts
```diff
@@ -0,0 +1,135 @@
...
+        }
+        // let json: any = {"failed": "failed"};
+        // json = new SyncRequestClient()
+        //     .post<object, object>('https://api.boavizta.org/v1/cloud/?verbose=true&allocation=TOTAL', data);
+        const response = await axios.post('https://api.boavizta.org/v1/cloud/', dataCast);
+        return response.data;
+    }
+}
+
+export class BoaviztaComponentImpactModel implements IImpactModelInterface {
```
jawache, line 70: This one looks the same as the other class, is it left here by mistake?

jawache (changes requested):

jawache:
Hey @gnanakeethan there is a lot here let's make this simple. I know previously asked to include the generated output but I think we've learned enough for now and adjusted the spec.

Let's close out this PR and create another one which includes just the TS code that we want to actually merge into the repo.

The PR we want to create is a PR with code that supports #49 only. I'll create another ticket or flesh that out specifically with what's required tonight.

lib/boavizta.ts
gnanakeethan: It will not be possible use "in" without casting.

lib/boavizta.ts
gnanakeethan: Nope. It is intentional.

gnanakeethan: It implements the component impact model which is distinct from cloud impact model.

gnanakeethan:
[Archive.zip](https://github.com/Green-Software-Foundation/carbon-ql/files/12157321/Archive.zip)


@srini1978 NuGet Packages

PR #60, opened 2023-07-18
Suggestions for Impact YAML Specification 

gnanakeethan:
I have included the changes as a YAML file itself. 

There is a schema generated using the YAML. 

1. Adds an optional field called "tags", and it can be used to isolate resources
2. Removes "params" from root of the node, since it does not provide any value.
3. Adds an optional field called "setup", it has fields & values as an array, it is used to setup the model. 
4. Updates the telemetry field, "metric", "unit" are required.  "data" or "path" and "mappings" can be included in telemetry. 
    1. mappings allow customization of the order of fields in a csv document.
    2. data should feature timestamp and duration of measurement and the value. 
    3. metric required always and they define which metric is being measured, eg: cpu, ram, water. 
    4. unit denotes the scale of metric or a percentage


gnanakeethan:
Example Model
```yaml
$schema: "./impact-yaml-schema.json"
name: sample-app
components:
  - name: app-server
    tags: # common tags for identifying the resource
      - name: model
        value: boavizta
        description: description
    model: # so we know which IMP to use to calculate this node
      path: com.boavizta.ComponentPlugin
      setup:
        - name: componentType
          value: cpu
        - name: name
          value: Intel Xeon Platinum 8272CL
        - name: core_units
          value: 2
        - name: location
          value: USA
    telemetry:
      metric: cpu
      unit: percentage
      data:
        - timestamp: 2020-01-01T00:00:00.000Z
          duration: 1h
          value: 18.392
  - name: db-server
    tags: # common tags for identifying the resource
      - name: model
        value: boavizta
        description: description
    model: # so we know which IMP to use to calculate this node
      path: com.boavizta.ComponentPlugin
      setup:
        - name: componentType
          value: cpu
        - name: name
          value: Intel SP8160
        - name: core_units
          value: 2
        - name: location
          value: USA
    telemetry:
      metric: cpu
      unit: percentage
      data:
        - timestamp: 2020-01-01T00:00:00.000Z
          duration: 1h
          value: 10
```

specs/impact-yaml-suggestion.yaml
```diff
@@ -0,0 +1,66 @@
+$schema: "./impact-yaml-schema.json"
+name: app-name # The name of the root node (the applicatio as a whole)
+components: # The nodes under this root node
+  - name: nodename # The name of this node
+    tags: # common tags for identifying the resource
```
jawache, line 5: I like the addition of tags

gnanakeethan: Need to explore the schema for that. The array syntax is easily parsable whereas parsing the dict could be a complicated scenario.

srini1978:
Looks good

jawache:
Test update

jawache:
@gnanakeethan note, I've merged what makes sense long term to this PR: https://github.com/Green-Software-Foundation/carbon-ql/pull/63 

This PR won't get merged and we'll close it out once the other gets merged, if there is a proposal to a change to the actual YAML it should be made to the spec that wat we can see what's being proposed we change (see #68 to see what I mean)

srini1978:
@jawache  So if we want to propose changes, are you suggesting that we put in our comments in #63 or put it in here in this PR ?

Issue #43, opened 2023-06-27
Using JSII for writing polyglot (multi-language) libraries


jawache:
### Discussed in https://github.com/Green-Software-Foundation/carbon-ql/discussions/41

<div type='discussions-op-text'>

<sup>Originally posted by **jawache** June 12, 2023</sup>
Was speaking to someone today and they mentioned that AWS use a tool called JSII to create their CSDK. 
https://github.com/aws/jsii

JSII lets you write code in TypeScript and then transpiles it into Python, Java and .Net, see: https://aws.amazon.com/fr/blogs/opensource/generate-python-java-dotnet-software-libraries-from-typescript-source/

I think it could be a good solution to write the initial set of libraries for carbon-ql.</div>

jawache:
We discussed last project call and @gnanakeethan will create some demo code using JSII so we can review on the call whether it might be a good initial framework to use.

gnanakeethan:
https://github.com/aws/jsii/issues/369

jawache:
Asim: AWS recommend this isn't used for high performance or long running processes, for those make language specific libraries. We would consider this mostly for single shot calculations.

gnanakeethan:
Unable to return a promise from JSII

jawache:
Thanks @gnanakeethan, re: "Using "object" as return type causes troubles in calling methods from the returned "object" in Typescript & Python." is the returned object from TypeScript translated into a normal Dict object in Python? Could you elaborate on what methods you want to call on object which do not function.

jawache:
Hi @gnanakeethan could you elaborate on the object question above?

PR #62, opened 2023-07-23
Update impact-model-plugin spec to reflect learnings from JSII and Boavizta

jawache:
Update the spec according to the findings in this issue https://github.com/Green-Software-Foundation/carbon-ql/issues/43 which investigated implementing the spec with Boavizta and JSII

1. Remove stream this is a feature of the thing that "calls" the plugin and the specific implementation will vary based on language/tech (e.g. threads, events). To stream this you just call calculate whenever your streaming implementation needs to.
2. Remove batch, this needs much more thought as per last project call (re: time, bucketing, synchronization), see https://github.com/Green-Software-Foundation/carbon-ql/issues/59
3. Update create -> configure, as per the learnings from #43 that we can't use the factory pattern.

### DIFF
--- a/specs/impact-model-plugin.md
+++ b/specs/impact-model-plugin.md
@@ -18,13 +18,15 @@ The goal of the Impact Measurement Plugin (IMP) Spec is to define a common stand
 
 ## Impact Model Interface Specification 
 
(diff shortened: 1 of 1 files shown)


jawache:
Agreed on call to merge.

PR #63, opened 2023-07-23
Update yaml to reflect spec update

jawache:
Updated the YAML spec to include learnings from 
https://github.com/Green-Software-Foundation/carbon-ql/pull/62
and 
https://github.com/Green-Software-Foundation/carbon-ql/pull/60

1. Added tags
2. Added clarity re: loading from CSV file.
3. Name change from telemetry -> observations to reflect that we are not consuming "raw" telemetry which is getting people in the SRE/DevOps space confused. We'll be inputting some already pre-aggregated measures, smallest duration will likely be 5 mins to match to the smallest duration of grid emissions data it's possible to gather. This isn't millisecond telemetry. In other contexts, e.g. time series data, a term for this is an observation.
4. Cleaned up re: use of arrays to list the components.

### DIFF
--- a/specs/impact-yaml.md
+++ b/specs/impact-yaml.md
@@ -42,14 +42,15 @@ components: # The nodes under this root node
       path: <a unique string, path to represent an IMP>
     config: # common static params
       <key>: <value>
-    measurement: # A single measurement
+    observation: # A set of data points for a moment and duration of time.
       <key>: <value>
       ...
-    measurements: # A series of measurements
+    observations: # Multiple sets of data points for multiples moments and durations of time.
       shared:
         <key>: <value> # Data that is repeated for every measurement, e.g machine type.
-      series: 
-    measurements: # A series of measurements stored in a CSV file
+      series:
+        - <first observation>
+    observations: # Observations stored in a CSV file
       path: <file> # Path to a CSV file containing the measurements
     components: # The sub nodes of this node
       - <name>: <another component same as above>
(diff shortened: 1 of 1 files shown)


srini1978 (changes requested):
Can we separate the duration field as 1) Duration and 2) Unit . This will allow us to scale the yaml in scenarios where telemetry is for a period of 2 hours or 3 hours rather than seconds

jawache:
@srini1978 I'm very keen for the default impl to be as minimal as possible, to make demo'ing, explaining it easier - 5 mins to wow.

So perhaps we can have a default duration units of "seconds" and if you want to provide durations in different units then that's extra configuration.

```yaml
    - datetime: 2023-07-06T00:05:
      duration: 5 # defaults to seconds
      cpu: 0.11
```

@gnanakeethan is proposing a mapping field for the CSV loading to convert/provide metadata for the observations so perhaps we can use that for even manually created

So perhaps something like this?

```yaml
series:
    - datetime: 2023-07-06T00:05:
      span: 5
      cpu: 11
mapping:
  span:
    units: minutes # field is in minutes
    to: duration # field needs to be converted to "duration" for input to the IMP.
  cpu:
    units: percentage
```

I like the idea of sticking to one mapping sections that works for both CSV and our static inputs. 

Thoughts?

jawache:
@srini1978 added the mapping.

jawache:
Agreed on project call

PR #70, opened 2023-07-28
Implementation for Boavizta Cloud Impact Model 

### LINKED ISSUE #68: Create Boavizta SCI Cloud IMP
## Name 
bovizta.cloud.sci

## Scope
This model returns the SCI score and component E, I and M values for cloud providers as supported by the Boavizta API

## References
- https://doc.api.boavizta.org/Reference/cloud_route/

## Assumptions
N/A

##  Limitations
- The Boavizta API currently only supports a provider of AWS for this model.

## Validations

- List of allowed country codes: https://api.boavizta.org/docs#/utils/utils_get_all_countries_v1_utils_country_code_get
- List of supported providers: https://api.boavizta.org/docs#/cloud/server_get_all_provider_name_v1_cloud_all_providers_get
- List of supported instance_types: https://api.boavizta.org/docs#/cloud/server_get_all_archetype_name_v1_cloud_all_instances_get

## Inputs
- provider: The name of the cloud provider
- instance_type: The instance type of the VM offering on that cloud provider.
- usage_location: The location where the VM is based
- days_use_time: Fraction of a day the VM is used for.
- time_workload: Average CPU utilization for that duration **or** a breakdown of the CPU utilization for that duration in this format:
  - Array of {
        "time_percentage": % of time,
        "load_percentage": utilization for this time slice
      },
      {
        "time_percentage": % of time,
        "load_percentage": utilization for this time slice
      }
- verbose: true/false
- allocation: Method of calcualtion, linear or total

## Example IMPL

```yaml
- component: 
    model: 
      path: bovizta.cloud.sci
    config: 
      verbose: true
      allocation: total
      provider: aws
      instance_type: r6g.medium
    observations:
      shared:
        location: FRA  # usage_location
      series:
          -   datetime: 2023-07-06T00:00
              duration: 15 # translated into days_use_time
              cpu: 0.34  # translated into time_workload                          
          -   datetime: 2023-07-06T00:15
              duration: 15s                              
              cpu: 0.12                                    
          -   datetime: 2023-07-06T00:30
              duration: 15s                               
              cpu: 0.01                                     
          -   datetime: 2023-07-06T00:40
              duration: 15s                               
              cpu: 0.78            
```

## Observations

- Will be in the format of CPU utilization, durations.
- They additionally have a location which is either used to give more accurate embodied or if it's used for grid emissions we can ignore since we want to handle grid emissions ourselves.

### Multiple Observations

```yaml
series:
    -   datetime: 2023-07-06T00:00
        duration: 15s                             
        cpu: 0.34                                    
    -   datetime: 2023-07-06T00:15
        duration: 15s                              
        cpu: 0.12                                    
    -   datetime: 2023-07-06T00:30
        duration: 15s                               
        cpu: 0.01                                     
    -   datetime: 2023-07-06T00:40
        duration: 15s                               
        cpu: 0.78                                     
```

- We need to call the API multiple times, **once** per observation.

**Eventually**, we will need to convert the Energy numbers that are returned into Carbon with our own Grid Emissions data source that provides real-time grid emissions.

```yaml
series:
    -   datetime: 2023-07-06T00:00
        e: 123 # convert this to O with the I for this specific time and duration using own own data source of grid emissions
        m: 34               
    -   datetime: 2023-07-06T00:15
        e: 12 # convert this to O with the I for this specific time and duration
        m: 34                           
    -   datetime: 2023-07-06T00:30
        e: 13 # convert this to O with the I for this specific time and duration
        m: 34                            
    -   datetime: 2023-07-06T00:40
        e: 234 # convert this to O with the I for this specific time and duration
        m: 34                                 
```

## Outputs

This is the raw output from Boavizta

```json
{
  "gwp": {
    "manufacture": 970, # embodied (m) in gCO2eq
    "use": 2100,
    "unit": "kgCO2eq"
  },
  "pe": {
    "manufacture": 13000,
    "use": 71020, # energy (e) in kWh
    "unit": "MJ"
  },
  "adp": {
    "manufacture": 0.15,
    "use": 0.000354,
    "unit": "kgSbeq"
  }
}
```

ImpactMetric (SCI style) will look like so for **now**:

```
{
 'e': xxx
 'm': xxx
}
```

## Unit Tests

- Calling with observations calls the API and returns a value for **e** and **m**.
- Call with a single observation returns a single impact metric with the same timestamp and duration.
- Call with multiple observations returns a multiple impact metrics with the same timestamp and duration as the original observation.
- If you don't provide mandatory inputs it returns an error (please decide what the mandatory inputs are in terms of config and observations).
- When the Boavizta API doesn't work (400/500 whatever) it returns an error.
- If you provide an instance type that isn't supported, returns an error
- If you provide a provider that isn't supported returns an error.
- If you provide a region that isn't supported returns an error.

gnanakeethan:
- Accepts array of objects or a single object as input.
- Transforms the input from IMPL to Boavizta compatible data type
- Allows raw input for ease of use.
- Exposes "configure", "calculate" functions as expected by the IMPCON.

Closes #68

### DIFF
--- a/lib/boavizta.ts
+++ b/lib/boavizta.ts
@@ -183,7 +183,7 @@ export class BoaviztaCloudImpactModel extends BoaviztaImpactModel implements IIm
             staticParams.allocation = undefined;
         }
         // if no valid provider found, throw error
-        this.validateProvider(staticParams);
+        await this.validateProvider(staticParams);
         // if no valid instance_type found, throw error
         await this.validateInstanceType(staticParams);
         // if no valid location found, throw error
@@ -222,11 +222,11 @@ export class BoaviztaCloudImpactModel extends BoaviztaImpactModel implements IIm
         }
     }
 
-    validateProvider(staticParamsCast: object) {
+    async validateProvider(staticParamsCast: object) {
         if (!('provider' in staticParamsCast)) {
             throw new Error("Improper configure: Missing provider parameter");
         } else {
-            const supportedProviders = ["aws"];
+            const supportedProviders = await this.supportedProvidersList();
             if (!(supportedProviders.includes(staticParamsCast.provider as string))) {
                 throw new Error("Improper configure: Invalid provider parameter: '" + staticParamsCast.provider + "'. Valid values are : " + supportedProviders.join(", "));
             }
@@ -236,6 +236,11 @@ export class BoaviztaCloudImpactModel extends BoaviztaImpactModel implements IIm
     async supportedInstancesList(provider: string) {
         const instances = await axios.get(`https://api.boavizta.org/v1/cloud/all_instances?provider=${provider}`)
         return instances.data;
+    }
+
+    async supportedProvidersList(): Promise<string[]> {
+        const providers = await axios.get(`https://api.boavizta.org/v1/utils/all_providers`)
+        return Object.values(providers.data);
     }
 
 
--- a/lib/boavizta.ts
+++ b/lib/boavizta.ts
@@ -156,3 +156,97 @@ export class BoaviztaCpuImpactModel extends BoaviztaImpactModel implements IImpa
     }
 }
 
+export class BoaviztaCloudImpactModel extends BoaviztaImpactModel implements IImpactModelInterface {
+    public sharedParams: object | undefined = undefined;
+    public instanceTypes: { [key: string]: string[] } = {};
+    public name: string | undefined;
+    public verbose: boolean = false;
+    public allocation: string = "TOTAL";
+
+    modelIdentifier(): string {
+        return "org.boavizta.cloud.sci"
+    }
+
+    protected async captureStaticParams(staticParams: object) {
+        const staticParamsCast = staticParams;
+        if ('verbose' in staticParamsCast) {
+            this.verbose = staticParamsCast.verbose as boolean ?? false;
+            staticParamsCast.verbose = undefined;
+        }
+        if ('allocation' in staticParamsCast) {
+            const allocation = staticParamsCast.allocation as string ?? "TOTAL";
+            // allocation has to be either TOTAL or LINEAR
+            if (["TOTAL", "LINEAR"].includes(allocation)) {
+                this.allocation = allocation;
+            } else {
+                throw new Error("Improper configure: Invalid allocation parameter. Either TOTAL or LINEAR");
+            }
+            staticParamsCast.allocation = undefined;
+        }
+        // if no valid provider found, throw error
+        this.validateProvider(staticParamsCast);
+        // if no valid instance_type found, throw error
+        await this.validateInstanceType(staticParamsCast);
+        // if no valid location found, throw error
+        await this.validateLocation(staticParamsCast);
+        this.sharedParams = Object.assign({}, staticParams);
+        return this.sharedParams
+    }
+
+    async validateLocation(staticParamsCast: object) {
+        if ('location' in staticParamsCast) {
+            const location = staticParamsCast.location as string ?? "USA";
+            const countries = await this.supportedLocations();
+            if (!(countries.includes(location))) {
+                throw new Error("Improper configure: Invalid location parameter: '" + location + "'. Valid values are : " + countries.join(", "));
+            }
+        }
+    }
+
+    async validateInstanceType(staticParamsCast: object) {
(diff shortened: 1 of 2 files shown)


lib/boavizta.ts
```diff
@@ -0,0 +1,283 @@
...
+    public name: string | undefined;
+    public verbose: boolean = false;
+    public allocation: string = "TOTAL";
+
+    modelIdentifier(): string {
+        return "org.boavizta.cloud.sci"
+    }
+
+    protected async captureStaticParams(staticParams: object) {
+        const staticParamsCast = staticParams;
```
jawache, line 202: What's the reason to assign to a new variable and use the new variable for checks?

lib/boavizta.ts
```diff
@@ -0,0 +1,283 @@
...
+        if ('verbose' in staticParamsCast) {
+            this.verbose = staticParamsCast.verbose as boolean ?? false;
+            staticParamsCast.verbose = undefined;
```
jawache, line 205: Why assign to undefined?

lib/boavizta.ts
```diff
@@ -0,0 +1,283 @@
...
+        }
+        if ('allocation' in staticParamsCast) {
+            const allocation = staticParamsCast.allocation as string ?? "TOTAL";
```
jawache, line 208: I'm not aware of this syntax, what does ?? do here?

lib/boavizta.ts
```diff
@@ -0,0 +1,283 @@
...
+        } else {
+            throw new Error("Improper configure: Missing instance_type parameter");
+        }
+    }
+
+    validateProvider(staticParamsCast: object) {
+        if (!('provider' in staticParamsCast)) {
+            throw new Error("Improper configure: Missing provider parameter");
+        } else {
+            const supportedProviders = ["aws"];
```
jawache, line 261: There is a Boavizta endpoint to return supported providers

lib/boavizta.ts
gnanakeethan: It was used when I had casted the input to index signature for easier working with key value pairs. I will remove.

lib/boavizta.ts
gnanakeethan: These two configuration do not go in as a part of JSON body rather a URL query parameter. Therefore it is removed after capturing the value.

lib/boavizta.ts
gnanakeethan: null coalescing operator. it returns the second value if the first is undefined or null.

lib/boavizta.ts
gnanakeethan: Updated.

jawache:
To fix the embodied issue let's add a configuration parameter called: expected-lifespan, with units of "years".
Then @gnanakeethan use the equation here https://github.com/Green-Software-Foundation/sci/blob/main/Software_Carbon_Intensity/Software_Carbon_Intensity_Specification.md#embodied-emissions to weight the embodied response from Boavizta's.

jawache:
@gnanakeethan after some investigation I think we just need to use the LINEAR allocation method.

Linear means that this output changes with respect to the amount of time you input, so if you input more time the manufacturing carbon emissions reported increases.

```
    "gwp": {
      "manufacture": 0.086,
      "use": 0.01,
      "unit": "kgCO2eq"
    },
```

jawache:
**Default**
For embodied carbon, use LINEAR and take manufacturing as is.
For energy default is to take it as is from the PE section.

**Configure It?**
- Perhaps explore doing our own embodied carbon attribution clacualtion from the raw data in here: https://api.boavizta.org/docs#/cloud/instance_cloud_impact_v1_cloud__post
- Perhaps returning secondary energy instead of primary energy by scaling the energy by some factor.

jawache (changes requested):
@gnanakeethan remove the python folder for cleanliness, this isn't needed (we're implementing in another ticket and in TS)

gnanakeethan:
@jawache I am unable to push to the branch

jawache (approved):
LGTM

jawache:
@gnanakeethan can you please fix the conflicts and we can merge already with pre-approved from the project team.

gnanakeethan:
@jawache Conflicts resolved via Rebasing.

gnanakeethan:
@jawache I am merging this since it is already approved.

PR #74, opened 2023-08-10
Cloud Carbon Foot Print Implementation

gnanakeethan:
Implements the CCF Version of the IEF. 

Provides the standard interface for calculating usage.


lib/ccf/index.ts
```diff
@@ -0,0 +1,259 @@
...
+        let wattage = 0;
+        if (this.provider === 'aws') {
+            const x = [0, 10, 50, 100];
+            const y: number[] = [
+                this.computeInstances['aws'][this.instanceType].consumption.idle ?? 0,
+                this.computeInstances['aws'][this.instanceType].consumption.tenPercent ?? 0,
+                this.computeInstances['aws'][this.instanceType].consumption.fiftyPercent ?? 0,
+                this.computeInstances['aws'][this.instanceType].consumption.hundredPercent ?? 0
+            ];
+            const spline = new Spline(x, y);
```
jawache, line 119: We should support different methods here. 
- Spline is one.
- Linear interpolation between points is another.
- Might have others?

lib/ccf/index.ts
```diff
@@ -0,0 +1,259 @@
...
+        // (wattage * duration) / (seconds in an hour) / 1000 = kWh
+        return ((wattage * duration) / 3600) / 1000;
+    }
+
+
+    modelIdentifier(): string {
+        return "ccf.cloud.sci";
+    }
+
+    standardizeInstanceMetrics() {
```
jawache, line 146: What is this doing? Can we add some detailed comments.

lib/ccf/index.ts
```diff
@@ -0,0 +1,259 @@
...
+                    "e": e,
+                    "m": m,
+                });
+            });
+
+        }
+        return results;
+    }
+
+    calculateEnergy(observation: { [key: string]: any; }) {
```
jawache, line 101: Let's add some comments.

lib/ccf/index.ts
```diff
@@ -0,0 +1,259 @@
...
+
+    constructor() {
+        this.standardizeInstanceMetrics();
+    }
+
+    authenticate(authParams: object): void {
+        this.authParams = authParams;
+    }
+
+    async configure(name: string, staticParams: object | undefined = undefined): Promise<IImpactModelInterface> {
```
jawache, line 51: Every configuration param needs to be documented in the commend above the function (this needs to be the same format for every IMP). 
Eventually we need this data in some external file format, people shouldn't have to read code to understand how to configure a YAML file. But for now documenting the settings in the commend block above configure is fine.

Makefile
jawache, line 1: Is this a file just for the ccf module? Is this only called manually or by the npm scripts when it installs this module? Can we add some comments at least so in the future we know why this script exists and when we can safely delete/move it.

Eventually the imps will be separate modules/repos, we won't be bundling the ccf, boavizta code etc... with the core ief framework, command line tool. So with the caveat happy to include this for the proof of concept, we will need to move this somewhere that sits where the ccf module will sit in the future.

jawache (commented):
This isn't approved by me I think @narekhovhannisyan and Cameron should take a look also but just some comments for now.

lib/ccf/index.ts
gnanakeethan: Linear Interpolation is used for GCP & Azure via the Architecture Type. 

As suggested by Cameron, I am now planning to use the hand-written list of instances from the CCF Repository for determining the architecture of an instance. 

We will lose the Spline interpolation for AWS once such is implemented. 

I still do carry the question whether the this [aws instances use](https://github.com/cloud-carbon-footprint/ccf-coefficients/blob/main/output/coefficients-aws-use.csv) data is for total instance usage or just for CPU ? Why there is a GB/Chip metric if not?

Makefile
gnanakeethan: This is only for the CCF. We can move it inside CCF but this result generated is needed for the CCF plugin to function.

gnanakeethan:
@jawache 

I am finding some conflict in the instance list vs the handwritten mapping. 

There are multiple architectures associated with a single instance family in GCP.  [Instance List](https://github.com/cloud-carbon-footprint/cloud-carbon-footprint/blob/43002ac7e6da2dabd33c656ce55babe04db91cf8/packages/gcp/src/lib/MachineTypes.ts#L31)


Can you please confirm if we should average all three and calculate further on?

lib/ccf/index.ts
gnanakeethan: I have now implemented linear interpolation using the handwritten definition in the ccf repository.

lib/ccf/index.ts
```diff
@@ -0,0 +1,378 @@
...
+    instanceType: string = '';
+    expectedLifespan = 4;
+
+    interpolation = Interpolation.LINEAR;
+
+    constructor() {
+        this.standardizeInstanceMetrics();
+    }
+
+    // Defined for compatibility. Not used in CCF.
```
narekhovhannisyan, line 61: We can use following format for all function documentations (JSDoc style documentation which can be used also for generating docs). It's possible to add references to docs, document params, description and so on.
```
/**
 * Defined for compatibility. Not used in CCF.
 */
authenticate(authParams: object): void {
        this.authParams = authParams;
}
```

lib/ccf/index.ts
```diff
@@ -0,0 +1,378 @@
...
+    *  Configuration Parameters for StaticParams
+    *
+    *  provider: aws, gcp, azure
+    *  instance_type: instance type from the list of supported instances
+    *  expected_lifespan: expected lifespan of the instance in years
+    *  interpolation: linear(All Clouds), spline (only for AWS)
+    */
+    async configure(name: string, staticParams: object | undefined = undefined): Promise<IImpactModelInterface> {
+        this.name = name;
+        if (staticParams === undefined) {
```
narekhovhannisyan, line 78: also before and after if statement you can keep new lines which makes function more readable

lib/ccf/index.ts
```diff
@@ -0,0 +1,378 @@
...
+            architecture = 'Average';
+        }
+        if (!(architecture in this.awsList)) {
+            console.log("ARCHITECTURE:", architecture)
+        }
+        return architecture;
+    }
+
+    // Calculates the embodied emissions for a given observation
+    embodiedEmissions(observation: { [key: string]: any; }): number {
```
narekhovhannisyan, line 361: in my PR I have introduced types for `{ [key: string]: any; }` this kind of things, so after merge we can reuse them in this code too.

gnanakeethan:
@narekhovhannisyan Please review. I have pushed and patched on top of dev.

src/lib/ccf/index.ts
```diff
@@ -0,0 +1,458 @@
...
+}
+
+export class CloudCarbonFootprint implements IImpactModelInterface {
+  // Defined for compatibility. Not used in CCF.
+  authParams: object | undefined;
+  // name of the data source
+  name: string | undefined;
+  // compute instances grouped by the provider with usage data
+  private computeInstances: {
+    [key: string]: {
```
narekhovhannisyan, line 51: custom types like this can be moved to src/types/cloud-carbon-footprint.ts file

src/lib/ccf/index.ts
```diff
@@ -0,0 +1,458 @@
...
+  }
+
+  /**
+   *  Configures the CCF Plugin for IEF
+   *  @param {string} name name of the resource
+   *  @param {Object} staticParams static parameters for the resource
+   *  @param {("aws"|"gcp"|"azure")} staticParams.provider aws, gcp, azure
+   *  @param {string} staticParams.instance_type instance type from the list of supported instances
+   *  @param {number} staticParams.expected_lifespan expected lifespan of the instance in years
+   *  @param {Interpolation} staticParams.interpolation linear(All Clouds), spline (only for AWS)
```
narekhovhannisyan, line 85: I think since we are using typescript types, we can skip param types and just describe them to not have conflict.
F. ex. in this case you have described the `staticParams` object in the doc, but its typescript type is `object` so it's better to focus on typescript types.

src/lib/ccf/index.ts
```diff
@@ -0,0 +1,458 @@
...
+   * Calculate the total emissions for a list of observations
+   *
+   * Each Observation require:
+   *  @param {Object[]} observations  ISO 8601 datetime string
+   *  @param {string} observations[].datetime ISO 8601 datetime string
+   *  @param {number} observations[].duration observation duration in seconds
+   *  @param {number} observations[].cpu percentage cpu usage
+   */
+  async calculate(
+    observations: object | object[] | undefined
```
narekhovhannisyan, line 148: if we know what kind of object is observation, then let's use that definition instead of object

src/lib/ccf/index.ts
```diff
@@ -0,0 +1,458 @@
...
+import * as gcp_instances from './gcp-instances.json';
+import * as gcp_use from './gcp-use.json';
+import * as aws_use from './aws-use.json';
+import * as azure_use from './azure-use.json';
+import * as azure_instances from './azure-instances.json';
+import * as gcp_embodied from './gcp-embodied.json';
+import * as aws_embodied from './aws-embodied.json';
+import * as azure_embodied from './azure-embodied.json';
+import {KeyValuePair} from '../../types/boavizta';
+
```
narekhovhannisyan, line 14: you can brake down imports part like this
```
import {INSTANCE_TYPE_COMPUTE_PROCESSOR_MAPPING} from '@cloud-carbon-footprint/aws/dist/lib/AWSInstanceTypes';
import Spline from 'typescript-cubic-spline';

import {IImpactModelInterface} from '../interfaces';

import * as aws_instances from './aws-instances.json';
import * as gcp_instances from './gcp-instances.json';
import * as gcp_use from './gcp-use.json';
import * as aws_use from './aws-use.json';
import * as azure_use from './azure-use.json';
import * as azure_instances from './azure-instances.json';
import * as gcp_embodied from './gcp-embodied.json';
import * as aws_embodied from './aws-embodied.json';
import * as azure_embodied from './azure-embodied.json';

import {KeyValuePair} from '../../types/boavizta';
```

to be logically separated like first goes native modules, then npm modules, interfaces, libs, utils, config, types

narekhovhannisyan: btw we can present json file imports in UPPERCASE manner to make them different from our local variables since when they are in lowercase, it seems like we are not keeping camelcase convention 

AWS_INSTANCES, GCP_INSTANCES, etc.

src/lib/ccf/index.ts
```diff
@@ -0,0 +1,458 @@
...
+export enum Interpolation {
+  LINEAR = 'linear',
+  SPLINE = 'spline',
+}
+
+export interface ICcfResult {
+  e: number;
+  m: number;
+}
```
narekhovhannisyan, line 42: all the types above can be moved to appropriate src/types/ccf.ts

src/lib/ccf/index.ts
```diff
@@ -0,0 +1,458 @@
...
+      observations.forEach((observation: KeyValuePair) => {
+        const e = this.calculateEnergy(observation);
+        const m = this.embodiedEmissions(observation);
+
+        results.push({
+          e: e,
+          m: m,
+        });
+      });
+    }
```
narekhovhannisyan, line 169: instead of declaring empty array then collecting, we can reduce obsercations array

```
return observations.reduce((observation) => {
  const calculatedObservation = {
    e: this.calculateEnergy(observation)
    m: this.embodiedEmissions(observation)
  }
  
  acc.push(calculatedObservation)
  
  return acc
}, [])
```

src/lib/ccf/index.ts
```diff
@@ -0,0 +1,458 @@
...
+    // Where:
+    // TE = Total Embodied Emissions, the sum of Life Cycle Assessment(LCA) emissions for all hardware components
+    // TR = Time Reserved, the length of time the hardware is reserved for use by the software
+    // EL = Expected Lifespan, the anticipated time that the equipment will be installed
+    // RR = Resources Reserved, the number of resources reserved for use by the software.
+    // TR = Total Resources, the total number of resources available.
+    const TotalEmissions =
+      this.computeInstances[this.provider][this.instanceType]
+        .embodiedEmission ?? 0;
+    const TimeReserved = duration_in_hours;
```
narekhovhannisyan, line 445: let's declare variables starting with lowercase since Uppercase first letter is associated with Classes

src/lib/ccf/index.ts
```diff
@@ -0,0 +1,458 @@
...
+
+    return architecture;
+  }
+
+  /**
+   * Calculates the embodied emissions for a given observation
+   */
+  private embodiedEmissions(observation: KeyValuePair): number {
+    // duration
+    const duration_in_hours = observation['duration'] / 3600;
```
narekhovhannisyan, line 434: durationInHours

src/lib/ccf/index.ts
```diff
@@ -0,0 +1,458 @@
...
+      this.computeInstances['azure'][instance['type']].embodiedEmission =
+        instance['total'];
+    });
+  }
+
+  // Architecture strings are different between Instances-Use.JSON and the bundled Typescript from CCF.
+  // This function resolves the differences.
+  private resolveAwsArchitecture(architecture: string) {
+    if (architecture.includes('AMD ')) {
+      architecture = architecture.substring(4);
```
narekhovhannisyan, line 401: after re-assigning I think all the other cases will not pass, so we can return to not execute other if statements

src/lib/ccf/index.ts
```diff
@@ -0,0 +1,458 @@
...
+  standardizeInstanceMetrics() {
+    this.computeInstances['aws'] = {};
+    this.computeInstances['gcp'] = {};
+    this.computeInstances['azure'] = {};
+    let gcpMin = 0.0;
+    let gcpMax = 0.0;
+    let gcpCount = 0;
+    // standardize gcp emissions
+    // gcp_use loaded from coefficients-gcp-use.csv file from CCF
+    gcp_use.forEach((instance: KeyValuePair) => {
```
narekhovhannisyan, line 264: This function can be simplified with breaking down into small functions, some of which can be reusable
if possible use reduce instead of for each for not declaring `sum variables` on each instance
```
    let gcpMin = 0.0;
    let gcpMax = 0.0;
    let gcpCount = 0;
```

src/lib/ccf/index.ts
gnanakeethan: @narekhovhannisyan the requirement is to realign the mismatched ones. 

There will be AMD Epyc and then we just need to cut off "AMD " from the string. 

Intel / Graviton will not come after that.

src/lib/ccf/index.ts
gnanakeethan: @narekhovhannisyan it is kept that way to avoid issues when compiling to different languages in JSII

src/lib/ccf/index.ts
gnanakeethan: it is kept that way to avoid issues when compiling to different languages in JSII

Sealjay:
I've been reading through the docs - is it possible for us to separate out the repository for third party plugins, data sources, and modules from the source code from IEF itself? Right now they look very coupled.

https://github.com/Green-Software-Foundation/ief/blob/dev/docs/Impact%20Model%20Plugin.md

jawache (approved):
Looks good t me, happy to merge as is.

PR #80, opened 2023-08-14
Adds impl spec and examples

jmcook1186:
Adds new directory containing a general specification document for IEF impl files and a set of examples. 
The examples translate the case studies from https://sci-guide.greensoftware.foundation/CaseStudies/ into the `impl` format so that they can become executable.

## Issue
Fixes https://github.com/Green-Software-Foundation/ief/issues/55#issuecomment-1675371216

### DIFF
--- a/impls/dow_msft.yaml
+++ b/impls/dow_msft.yaml
@@ -1,0 +1,84 @@
+name: dow-msft
+description: "https://github.com/Green-Software-Foundation/sci-guide/blob/dev/use-case-submissions/dow-msft-Graph-DB.md"
+tags:
+  kind: db-api
+  complexity: simple
+  category: cloud
+config:
+  pipeline:
+    calculation: dow-msft
+    normalization: default
+    aggregation: default
+# Single observation
+graph:
+  backend: # an advanced grouping node
+    model: default  
+    config: 
+      vendor: azure
+      region: east-us  # lookup carbon intensity from region
+    children: 
+      tiger-database: # a leaf component
+        observations: 
+          common:
+            n_cpu: 8
+            ram: 32
+            server: Intel-xeon-platinum-8380 # will lookup TPU and total embodied coefficients from model name
+          series:
+            - timestamp: 2023-07-06T00:00
+              span: 1 # this data is using span, but the model expects duration
+              tdp_coeff: 0.4 #% total available cpu usage as fraction in range 0-1
+              cpu_utilization: 17.12
+              memory_utilization: 6.2
+          mapping:
+            span:
+              units: hours
+              to: duration
+      tiger-api: # a leaf component
+        observations: 
+          common:
+            n_cpu: 1
+            ram: 1.75 #total allocated
+            server: Intel-xeon-platinum-8270 # will lookup TPU and total embodied coefficients from model name
+          series:      
+            - datetime: 2023-08-06T00:00
+              duration: 1
+              tdp_coeff: 0.4
+              cpu_utilization: 25
+              memory_utilization: 1.22 # 70% of memory allocation, as per docs
+          mapping:
+            span:
+              units: hours
+              to: duration
+      neo4j-database: # a leaf component
+        observations: 
+          common:
+            n_cpu: 8
+            ram: 32 # total allocated in GB
+            server: Intel-xeon-platinum-8380 # will lookup TPU and total embodied coefficients from model name
+          series:
+            - timestamp: 2023-07-06T00:00
+              span: 1 # this data is using span, but the model expects duration
+              tdp_coeff: 0.5 #% total available cpu usage as fraction in range 0-1
+              cpu_utilization: 28.05
+              memory_utilization: 6.3
+          mapping:
+            span:
+              units: hours
+              to: duration
+      neo4j-api: # a leaf component
+        observations: 
+          common:
+            n_cpu: 1
+            ram: 1.75 #total allocated
+            server: Intel-xeon-platinum-8270 # will lookup TPU and total embodied coefficients from model name
+          series:      
+            - datetime: 2023-08-06T00:00
+              duration: 1
+              tdp_coeff: 0.8
+              cpu_utilization: 14
+              memory_utilization: 1.13 # 65% of memory allocation, as per docs
+          mapping:
+            span:
+              units: hours
+              to: duration
+
--- a/impls/spec.md
+++ b/impls/spec.md
@@ -34,3 +34,93 @@ graph: # [KEYWORD] [ANY-SUBFIELDS] graph is the set of components that make up t
 
 # ... more children 
 ```
+
+
+## Example
+
+```yaml
+name: dow-msft
+description: "https://github.com/Green-Software-Foundation/sci-guide/blob/dev/use-case-submissions/dow-msft-Graph-DB.md"
+tags:
+  kind: db-api
+  complexity: simple
+  category: cloud
+config:
+  pipeline:
+    calculation: dow-msft
+    normalization: default
+    aggregation: default
+# Single observation
+graph:
+  backend: # an advanced grouping node
+    model: default  
+    config: 
+      vendor: azure
+      region: east-us  # lookup carbon intensity from region
+    children: 
+      tiger-database: # a leaf component
+        observations: 
+          common:
+            n_cpu: 8
+            ram: 32
+            server: Intel-xeon-platinum-8380 # will lookup TPU and total embodied coefficients from model name
+          series:
+            - timestamp: 2023-07-06T00:00
+              span: 1 # this data is using span, but the model expects duration
+              tdp_coeff: 0.4 #% total available cpu usage as fraction in range 0-1
+              cpu_utilization: 17.12
+              memory_utilization: 6.2
+          mapping:
+            span:
+              units: hours
+              to: duration
+      tiger-api: # a leaf component
+        observations: 
+          common:
+            n_cpu: 1
+            ram: 1.75 #total allocated
+            server: Intel-xeon-platinum-8270 # will lookup TPU and total embodied coefficients from model name
+          series:      
+            - datetime: 2023-08-06T00:00
+              duration: 1
+              tdp_coeff: 0.4
+              cpu_utilization: 25
+              memory_utilization: 1.22 # 70% of memory allocation, as per docs
+          mapping:
+            span:
+              units: hours
+              to: duration
+      neo4j-database: # a leaf component
+        observations: 
+          common:
+            n_cpu: 8
+            ram: 32 # total allocated in GB
+            server: Intel-xeon-platinum-8380 # will lookup TPU and total embodied coefficients from model name
+          series:
+            - timestamp: 2023-07-06T00:00
+              span: 1 # this data is using span, but the model expects duration
+              tdp_coeff: 0.5 #% total available cpu usage as fraction in range 0-1
+              cpu_utilization: 28.05
+              memory_utilization: 6.3
+          mapping:
+            span:
+              units: hours
+              to: duration
+      neo4j-api: # a leaf component
+        observations: 
+          common:
+            n_cpu: 1
+            ram: 1.75 #total allocated
+            server: Intel-xeon-platinum-8270 # will lookup TPU and total embodied coefficients from model name
+          series:      
+            - datetime: 2023-08-06T00:00
+              duration: 1
+              tdp_coeff: 0.8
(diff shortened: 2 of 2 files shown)


impls/msft-eshoppen.yaml
```diff
@@ -0,0 +1,96 @@
...
+      embodied_c: 350 # kgCO2eq
+      reserved_time: 1 # hour
+      resources_reserved: 1
+      total_resources_available: 1
+      lifespan: 4 #years
+    children:
+      cpu:
+        observations:
+          common:
+            server:  Intel® Xeon® Platinum 8272CL
```
jawache, line 25: Some thoughts, if the server isn't required for the actual lookup/calculation then maybe we should add it to to a tags node.

impls/msft-eshoppen.yaml
```diff
@@ -0,0 +1,96 @@
...
+            tdp: 28 # W
+            carbon_intensity: 951
+            tdp_coeff: 0.32
```
jawache, line 28: I see how this works now, it's pulling the TD coefficient using the utilization. We should represent this as a "teads" or some sort of generic "lookup" model, will discuss more in an issue.

impls/msft-eshoppen.yaml
```diff
@@ -0,0 +1,96 @@
...
+      resources_reserved: 1
+      total_resources_available: 1
+      lifespan: 4 #years
+    children:
+      cpu:
+        observations:
+          common:
+            server:  Intel® Xeon® Platinum 8272CL
+            tdp: 28 # W
+            carbon_intensity: 951
```
jawache, line 27: Starting to think that grid emissions is just another impact metric and should be treated the same as energy or embodied. As in there is a model that has observations input into it and outputs grid emissions.

If we are going to use the same grid emissions for all calculations then we can just configure a static model on the root node that returns the same value no matter what!

impls/ntt-data-on-premise.yaml
```diff
@@ -0,0 +1,165 @@
...
+        common:
+            p_baseline: 92.5 # W Baseline power based on full port 100% traffic for Cisco Catalyst 9300
+            n_ports: 24
+            c_intensity: 457 # gCO2/kwh
+            embodied_c: 0.251 #g per unit
+            lifespan: 4 # years
+            pue: 1.5
+        observations: 
+          series:
+            timestamp: 2023-07-06T00:00
```
jawache, line 29:
```suggestion
             - timestamp: 2023-07-06T00:00
```

impls/ntt-data-on-premise.yaml
```diff
@@ -0,0 +1,165 @@
...
+    enrichment: default
+# Single observation
+graph:
+  layer-3-switch: # an advanced grouping node
+    model: default  
+    config:
+    children: 
+      switch-1:
+        common:
+            p_baseline: 92.5 # W Baseline power based on full port 100% traffic for Cisco Catalyst 9300
```
jawache, line 21: Speaking to my previous point I think we have 3 models here.

- A new energy model for network switches: that is using a custom equation for calculating the energy for a network interface, takes as input baseline and observations of input/output rates.
- PUE model: Somehow scales the energy value (so there is a hierarchy of models?)
- sci embodied model, same as all the previous use cases.

impls/ntt-data-on-premise.yaml
```diff
@@ -0,0 +1,165 @@
...
+        common:
+          c_intensity: 457 # gCO2/kwh
+          embodied_c: 6270 #g per unit
+          lifespan: 4 # years
+          pue: 1.5
+        observations: 
+        series:
+          timestamp: 2023-07-06T00:00
+          duration: 1
+          instantaneous-power: 176
```
jawache, line 117: - There is an embodied model here same as the other examples.
- There is a different energy model used here though, this takes as input observations of average power during the observation interval and outputs a kWh (energy value). So effectively a simple power (kW) -> energy (kWh) model..

jawache (commented):
Hey @jmcook1186 great work. I've put lots of comments in but overall this has been a great exercise in understanding how the IEF could be used to calculate the current case studies.

I've realized our current approach won't cut it, what we need is multiple models per component/node. Will write that up as an issue, if you can keep my larger comments unresolved here I want to convert them into an issue.

jawache:
@jmcook1186 add Accenture's Case Study to the IMPLs.

jawache:
Hi @jmcook1186 I've updated just one yaml according to the new model pipeline spec, see this one: https://github.com/Green-Software-Foundation/ief/blob/cbb86245c750b8bc666e8725a5a5887da1a47580/impls/msft-eshoppen.yaml

See the updated spec here: https://github.com/Green-Software-Foundation/ief/blob/9185795d31178093962ed71c5f3c3fb2cc8a946c/docs/Model%20Pipeline.md

Also to make life easier for now I just defaulted everything to seconds.

I'm hoping from this one you'll get a reasonable idea of the approach and be able to apply it to others.

impls/accenture.yaml
```diff
@@ -0,0 +1,175 @@
...
+          duration: 2419200 # seconds in a month (7 days * 4 weeks)
+          cpu_utilization: 20
+          ram_utilization: 0
+          power: 5.70
+          energy: 4.62
+      mapping: 
+          units: month
+  app-gateway: # note that the app gateway does not have observations, its SCI is assumed equal to 5% of the total emissions
+    config:
+      sci: 1053.94
```
narekhovhannisyan, line 175: please add newline at the end of file

impls/farm-insights.yaml
```diff
@@ -0,0 +1,21 @@
...
+    model: sci-coefficient
+    config:
+      e: 0.003
+      i: 0.713
+      r: 86400 # 1 day in seconds
+      te: 55
+      tir: 86400 # 1 day in seconds
+      el: 94608000 # 1095 days in seconds
+      rr: 1
+      tor: 1
```
narekhovhannisyan, line 21: same here

impls/msft-eshoppen.yaml
```diff
@@ -0,0 +1,87 @@
...
+        i: 951
+      teads-curve:
+        processor:  Intel® Xeon® Platinum 8160
+        tdp: 150 # W
+    observations:
+      - timestamp: 2023-07-06T00:00 # [KEYWORD] [NO-SUBFIELDS] time when measurement occurred
+        duration: 3600
+        cpu: 10
+      - mapping:
+        units: hour
```
narekhovhannisyan, line 87: same here

impls/msft-green-ai.yaml
```diff
@@ -0,0 +1,44 @@
...
+        processor: nvidia-t4-nc16as-v3
+        e: 14.37
+        i: 1 # data not provided in case study
+        m: 0
+    observations: 
+      series:
+        - timestamp: 2023-07-06T00:00
+          duration: 1205 # 22.27 min runtime in seconds
+      mapping:
+        units: duration # R is per job, and the job took `duration` seconds
```
narekhovhannisyan, line 44: same here

impls/msft-green-ai.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - name: sci-c # a model that given e, i and m calculates a carbon value (e * i) + m
+      path: builtin       
+graph:
+  DenseNet: # an advanced grouping node
+    pipeline: 
+      - sci-c # energy & grid-carbon-intensity & embodied -> carbon
+    config: 
+      sci-c:
+        processor: nvidia-t4-nc16as-v3
+        e: 17.97
```
jawache, line 21: This is an observation I would say, the core one (GPUs are the only things that actually report energy directly)
Add to the observation as "energy"

impls/msft-green-ai.yaml
```diff
@@ -0,0 +1,44 @@
...
+        i: 1 # data not provided in case study
+        m: 0
```
jawache, line 23: I would add this an ab observation as well for simplicity, see my "add-obs" model in accenture.yml for an example of how we might easily add repeated values to Observations.

impls/msft-green-ai.yaml
```diff
@@ -0,0 +1,44 @@
...
+      path: builtin       
+graph:
+  DenseNet: # an advanced grouping node
+    pipeline: 
+      - sci-c # energy & grid-carbon-intensity & embodied -> carbon
+    config: 
+      sci-c:
+        processor: nvidia-t4-nc16as-v3
+        e: 17.97
+        i: 1 # data not provided in case study
```
jawache, line 22: Again, see add-obs from accenture.yml

impls/msft-green-ai.yaml
```diff
@@ -0,0 +1,44 @@
+name: mft-green-ai
```
jawache, line 1: This one is impossible to model I think TBH, not in a way that exactly mirrors the case study. Too much data is missing from the case study. But we can play around and have a think about ways to make this interesting.

impls/msft-green-ai.yaml
```diff
@@ -0,0 +1,44 @@
...
+description: "https://github.com/Green-Software-Foundation/sci-guide/blob/dev/use-case-submissions/msft-green-ai.md"
+tags:
+  kind: ml
+  complexity: simple
+  category: cloud
+config:
+  funit:
```
jawache, line 8: Can see where you are heading here but we can handle this with a model that just divides by the funit and therefore this is just another model with config.

The sci model will divide the `carbon` value with anything else that looks like `r` and pump the result out as a value called `sci`

impls/msft-green-ai.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - name: job
+      scale-factor: 1 # divide SCI by this value to get SCI per ml job
+  models:
+    - name: sci-c # a model that given e, i and m calculates a carbon value (e * i) + m
```
jawache, line 12: I think we can use a model called sci-o (operational emissions, just energy * grid-ci), still a carbon value but if we have no m then this might be enough.

impls/msft-green-ai.yaml
```diff
@@ -0,0 +1,44 @@
...
+description: "https://github.com/Green-Software-Foundation/sci-guide/blob/dev/use-case-submissions/msft-green-ai.md"
+tags:
+  kind: ml
+  complexity: simple
+  category: cloud
+config:
+  funit:
+    - name: job
+      scale-factor: 1 # divide SCI by this value to get SCI per ml job
+  models:
```
jawache, line 11: The case study talks about using watttime data. I think we can imagine here also including a model called `watttime` that calls the wattime service for a particual time and region to get the grid-ci value to add to the observation.

impls/msft-green-ai.yaml
```diff
@@ -0,0 +1,44 @@
...
+      sci-c:
+        processor: nvidia-t4-nc16as-v3
+        e: 14.37
+        i: 1 # data not provided in case study
+        m: 0
+    observations: 
+      series:
+        - timestamp: 2023-07-06T00:00
+          duration: 1205 # 22.27 min runtime in seconds
+      mapping:
```
jawache, line 43: We don't need these mappings any more, at least for now. 
Let's just assume every value is in the default units for that field (duration is seconds for instance). it makes like for us easier for us.
We can add a mapping functionality later on, or man, it could just be another model again with it's own config!

jawache (changes requested):
@jmcook1186 we're getting much closer to the solution here. I adjusted accentute.yml to how I would configure it and left a few comments for the others. 

Still thin it needs more refinement but we are close!

impls/msft-green-ai.yaml
jmcook1186: yes, I just included a skeleton for completeness - agree it's not very useful yet.

impls/msft-green-ai.yaml
jmcook1186: yes, this makes sense

impls/gsf-website.yaml
```diff
@@ -0,0 +1,93 @@
...
+    config:
+      co2js:
+        kind: 1byte # https://developers.thegreenwebfoundation.org/co2js/models/
+        green-hosting: true
+      sci:
+        time: month
+    observations:
+      - timestamp: 2023-07-06T00:00 
+        data: 8.267195767195768e-06 # 20 GB data transferred in most recent month, converted to GB/s
+        users: 2000 0.0008267195767195768 # 2000 unique users in most recent month, converted to users/s
```
narekhovhannisyan, line 93: add newline at the end of file

PR #95, opened 2023-08-25
Updated spec with model pipelines

jawache:
### Types of changes

- Updated the spec to explain model plugin pipelines.

### DIFF
--- a/docs/Model Pipeline.md
+++ b/docs/Model Pipeline.md
@@ -34,13 +34,15 @@ All the model plugins used in any component in the graph are configured at the t
 ```yaml
 .
 .
-initialize:
-  plugins:
+config:
+  models:
     - name: <model-name-1>
+      kind: builtin | plugin | shell
       path: <path-to-ts-module-to-load>
       config:
 	       <key>: <value>
     - name: <model-name-2>
+      kind: builtin | plugin | shell    
       path: <path-to-ts-module-to-load>
       config:
 	       <key>: <value>	       
@@ -55,7 +57,10 @@ In component nodes, we then configure the models we want to use in a pipeline li
 
 ```yaml
 backend:
-  pipeline: "model-1 | model-2 | model-3"
+  pipeline:
+    - model-1
+    - model-2
+    - model-3  
   config: 
     model-1:
       <key>: <value>
@@ -80,27 +85,17 @@ backend:
 - `config` in this part of the graph is config for each model in the pipeline (if any is required). Since we have multiple models, we need to define each config independently.
 - `observations` are the source observations, the values we are pumping into the start of this pipeline.
 
-The above is the short form; the long form would be to use a YAML array, like so:
 
-```yaml
-backend:
-  pipeline: 
-    - model-1
-    - model-2
-    - model-3
-  config: 
-    model-1:
-      <key>: <value>
-    model-2:
-      <key>: <value>        
-    model-3:
-      <key>: <value>
-```
 ## Example
 
 Let's look at a simple pseudo example, to begin with a pipeline like so:
 
-`instance-metadata | tdp | teads-curve`
+```yaml
+pipeline:
+  - instance-metadata
+  - tdp
+  - teads-curve
+```
 
 Using the above, we can combine multiple smaller models together to calculate the energy consumed by this observation:
 
@@ -108,9 +103,9 @@ Using the above, we can combine multiple smaller models together to calculate th
 observations: 
   - timestamp: 2023-07-06T00:00
     vendor: aws
-    instance_type: m5d.large
+    instance-type: m5d.large
     duration: 5
-    cpu: 33
+    cpu-util: 33
 ```
 
 ### `instance-metadata`
@@ -121,9 +116,9 @@ This model plugin takes as input an *instance type* and outputs the name of the 
 observations: 
   - timestamp: 2023-07-06T00:00
     vendor: aws
-    instance_type: m5d.large
+    instance-type: m5d.large
     duration: 5
-    cpu: 33
+    cpu-util: 33
 ```
 
 to 
@@ -132,12 +127,12 @@ to
 observations: 
   - timestamp: 2023-07-06T00:00
     vendor: aws
-    instance_type: m5d.large
+    instance-type: m5d.large
     duration: 5
-    cpu: 33
-    physical_processor: Intel Xeon Platinum 8175 # <-- output
-    used_cores: 1 # <-- output
-    total_cores: 26	# <-- output  	 
+    cpu-util: 33
+    physical-processor: Intel Xeon Platinum 8175 # <-- output
+    used-cores: 1 # <-- output
+    total-cores: 26	# <-- output  	 
 ```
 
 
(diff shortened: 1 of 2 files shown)


narekhovhannisyan (approved):

docs/Model Pipeline.md
```diff
@@ -0,0 +1,351 @@
...
+
+- `pipeline` defines the models we apply and the order in which we use them.
+- `config` in this part of the graph is config for each model in the pipeline (if any is required). Since we have multiple models, we need to define each config independently.
+- `observations` are the source observations, the values we are pumping into the start of this pipeline.
+
+The above is the short form; the long form would be to use a YAML array, like so:
+
+```yaml
+backend:
+  pipeline:
```
gnanakeethan, line 87: it will be better to maintain a single format.

jmcook1186 (approved):
LGTM. I wonder in some cases whether the distinction between chaining models and simply cleaning/preprocessing data from the impl might get a bit blurry? Probably just semantics.

docs/Model Pipeline.md
```diff
@@ -0,0 +1,351 @@
...
+    - timestamp: 2023-07-06T00:05
+      duration: 5
+      cpu: 23
+    - timestamp: 2023-07-06T00:10
+      duration: 5
+      cpu: 11
+```
+
+
+- `pipeline` defines the models we apply and the order in which we use them.
```
jmcook1186, line 79: do the time normalization, aggregation etc count as `models` in this spec, or are they defined separately? In the old spec they were defined in the `pipeline` field.

docs/Model Pipeline.md
```diff
@@ -0,0 +1,351 @@
...
+```
+
+## Usage
+
+In component nodes, we then configure the models we want to use in a pipeline like so:
+
+```yaml
+backend:
+  pipeline: "model-1 | model-2 | model-3"
+  config:
```
jmcook1186, line 59: maybe here we can define the functional unit and the operation required to represent the SCI in those units. For example, several of the current case studies calculate the SCI per s/hour/month and then retroactively apply some scaling to transform it into a SCI per API request or per website visit. This could be captured as follows:

```
config:
    funit:
        - name: request
          scale-factor: 32000 # divide SCI (per hour) by this value to get SCI per api call
```

jawache: Yeah I pulled back from describing the sci approach using a pipeline. It gets complex! But let's add this context in another doc just about how to implement the SCI

docs/Model Pipeline.md
jawache: Ok good point, will update to one format

docs/Model Pipeline.md
```diff
@@ -0,0 +1,351 @@
...
+- `pipeline` defines the models we apply and the order in which we use them.
+- `config` in this part of the graph is config for each model in the pipeline (if any is required). Since we have multiple models, we need to define each config independently.
+- `observations` are the source observations, the values we are pumping into the start of this pipeline.
+
+The above is the short form; the long form would be to use a YAML array, like so:
+
+```yaml
+backend:
+  pipeline: 
+    - model-1
```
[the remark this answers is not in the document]
jawache: Oh hmm i see what you're saying. So like:

pipeline:
```
  - iee:
    type: external
    path: python pimpl.py --imp <path-to-imp-file>
```

You could have types:

plugin (a ts module you dynamically load)
builtin (a backed in model, probably things around data manipulation)
shell (a shell command - both plugin and external are "external things" so we need something else to denote what's goin on).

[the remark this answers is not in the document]
jawache: added to spec

jawache:
> LGTM. I wonder in some cases whether the distinction between chaining models and simply cleaning/preprocessing data from the impl might get a bit blurry? Probably just semantics.

That's a good point. I've been calling everything a model. I think it's worthwhile still calling them models but having the understanding some will be simple processors?

docs/Model Pipeline.md
jawache: time-normalization can I realize be implemented as model like everything else (needs slight tweaking to the model initialization rules but it can work). Aggregation also now is just a normal standard aggregation and can just be baked into the framework.

docs/Model Pipeline.md
```diff
@@ -0,0 +1,364 @@
...
+
+to 
+
+```yaml
+observations: 
+  - timestamp: 2023-07-06T00:00
+    vendor: aws
+    instance-type: m5d.large
+    duration: 5
+    cpu-util: 33
```
gnanakeethan, line 132: Having hyphens in middle causes the parsers to behave differently and especially in javascript, the object access syntax we use will be void and we need to use the array syntax.

docs/Model Pipeline.md
```diff
@@ -0,0 +1,364 @@
...
+    - aws-lambda-to-instance 
+    - instance-metadata
+    - tdp
+    - teads      
+  observations: 
+    - timestamp: 2023-07-06T00:00
+      duration: 5
+      gb-s: 1005
+      vendor: aws # <-- new
+      instance_type: m5d.large  # <-- new
```
gnanakeethan, line 344: this declaration is with underscores.

srini1978:
> > LGTM. I wonder in some cases whether the distinction between chaining models and simply cleaning/preprocessing data from the impl might get a bit blurry? Probably just semantics.
> 
> That's a good point. I've been calling everything a model. I think it's worthwhile still calling them models but having the understanding some will be simple processors?

The distinction between the 2 is : there is a "component" which enriches the observation(s) and passes it on to the next model in the pipeline and the second is a component which actually calculates an Impact metric. In the true sense if we want to brand impact plugin as a software component that calculates impact metrices only then do we want to call these as ModelData Enrichment components. Here I am referring to our Observation as Model Data.

docs/Model Pipeline.md
```diff
@@ -0,0 +1,364 @@
...
+      to-instance: m5d.xlarge            
+  observations: 
+    - timestamp: 2023-07-06T00:00
+      duration: 5
+      cpu: 17.5 # <-- updated
+      vendor: aws
+      instance_type: m5d.xlarge # <-- updated
+```
+
+`m5d.xlarge` is the same CPU but twice the size of `m5d.large`, so this plugin halves the utilization to mirror what might be on the new instance type.
```
srini1978, line 294: Here the utilization is an observation. It needs to come from the software only. What we could do is by changing the instance type using this simulation the impact model can output sim-m which is the embodied emissions change.

jawache:
To review @srini1978 comments offline and confirm if there are any updates/objections before merging.

docs/Model Pipeline.md
jawache: There is still lots we need to think through around this, eventually I think there will form a dictionary of model terms.

Extract Model
Simulation Model
Adapter Model

I don't think there is any need to figure this out now but over time I think we'll evolve a language around this but it will be determined by the types of models that will be created.

RE: The confusion regarding **impacts** as a term for something that doesn't match our original defn of impacts. I agree but I'm not confident we have enough information yet to decide on an alternative approach? I suspect once we create more models and IMPLs then even more nuance might appear and an alternative language might surface.

Shall we aim to review this language around the decarb event and maybe our alpha release?

jawache: Moving this discussion to an issue since I think it's not blocking the merging of this spec. https://github.com/Green-Software-Foundation/ief/issues/149

PR #105, opened 2023-08-28
Load impl as cli param

narekhovhannisyan:
### Types of changes

- Initialized script for pipeline computation.


src/util/yaml.ts
```diff
@@ -0,0 +1,38 @@
+import {readFile, writeFile} from 'fs/promises';
+
+import * as YAML from 'js-yaml';
+
+import {ImpactYaml} from '../types/yaml';
```
jawache, line 5: @narekhovhannisyan it's challenging to type the yaml right now, everything under graph is not typed and also under the top level config/initilize node is also not typed. Both those sections are completely free form, every impl will be different.

There might be more we can type later on but right now super hard to judge, keep things untyped for as long as possible.

src/types/yaml.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export interface ImpactYaml {
+  name: string;
+  description: string;
+  tags: Record<string, string>;
+  config: Config;
+  enrichment: EnrichmentPipeline;
+  normalization: NormalizationPipeline;
+  aggregation: AggregationPipeline;
+  graph: Graph;
```
jawache, line 79: Make this a nested object/dictionary - everything is variable under here.

src/types/yaml.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export interface ImpactYaml {
+  name: string;
+  description: string;
+  tags: Record<string, string>;
+  config: Config;
+  enrichment: EnrichmentPipeline;
+  normalization: NormalizationPipeline;
+  aggregation: AggregationPipeline;
```
jawache, line 78: Not needed anymore

src/types/yaml.ts
```diff
@@ -0,0 +1,80 @@
...
+  functional_unit: string;
+}
+
+export interface ImpactYaml {
+  name: string;
+  description: string;
+  tags: Record<string, string>;
+  config: Config;
+  enrichment: EnrichmentPipeline;
+  normalization: NormalizationPipeline;
```
jawache, line 77: Not needed anymore

src/types/yaml.ts
```diff
@@ -0,0 +1,80 @@
...
+interface AggregationPipeline extends Pipeline {
+  functional_unit: string;
+}
+
+export interface ImpactYaml {
+  name: string;
+  description: string;
+  tags: Record<string, string>;
+  config: Config;
+  enrichment: EnrichmentPipeline;
```
jawache, line 76: Not needed anymore

src/types/yaml.ts
```diff
@@ -0,0 +1,80 @@
...
+
+interface AggregationPipeline extends Pipeline {
+  functional_unit: string;
+}
+
+export interface ImpactYaml {
+  name: string;
+  description: string;
+  tags: Record<string, string>;
+  config: Config;
```
jawache, line 75: Again, unsure exactly what this will contain. There will be some content that is typeable but for instance this is where we will configure the plugins, and we will never know the parameters that plugins will have in their config so there is a level here where the config will just need  be [key]: any

jawache (changes requested):
@narekhovhannisyan looks good overally, but we need to scale back the type conformity for the yaml as per my comments. You'll see as you start looking at more impl files it's a free form tree structure and most of the type information is in plugins which eventually will be dynamically imported so not something we want in our core framework.

jawache:
@narekhovhannisyan looks good to me happy to merge

PR #107, opened 2023-08-30
Teads Curve Model

gnanakeethan:
### Types of changes

### Teads CPU Implementation based on the power curve published at [medium.com/teads-engineering/building-an-aws-ec2-carbon-emissions-dataset-3f0fd76c98ac](https://medium.com/teads-engineering/building-an-aws-ec2-carbon-emissions-dataset-3f0fd76c98ac)

### DIFF
--- a/src/lib/teads-curve/index.ts
+++ b/src/lib/teads-curve/index.ts
@@ -2,7 +2,7 @@ import {IImpactModelInterface} from '../interfaces';
 import Spline from 'typescript-cubic-spline';
 import {KeyValuePair} from '../../types/boavizta';
 
-export class TeadsCPUModel implements IImpactModelInterface {
+export class TeadsCurveModel implements IImpactModelInterface {
   // Defined for compatibility. Not used in TEADS.
   authParams: object | undefined;
   // name of the data source
--- a/src/lib/teads-curve/index.ts
+++ b/src/lib/teads-curve/index.ts
@@ -77,8 +77,8 @@ export class TeadsCurveModel implements IImpactModelInterface {
    * Calculate the total emissions for a list of observations
    *
    * Each Observation require:
-   *  @param {Object[]} observations  ISO 8601 datetime string
-   *  @param {string} observations[].datetime ISO 8601 datetime string
+   *  @param {Object[]} observations  RFC3339 datetime string
+   *  @param {string} observations[].datetime RFC3339 datetime string
    *  @param {number} observations[].duration observation duration in seconds
    *  @param {number} observations[].cpu percentage cpu usage
    */
@@ -109,7 +109,7 @@ export class TeadsCurveModel implements IImpactModelInterface {
    *
    * duration: duration of the observation in seconds
    * cpu: cpu usage in percentage
-   * datetime: ISO 8601 datetime string
+   * datetime: RFC3339 datetime string
    *
    * Uses a spline method on the teads cpu wattage data
    */
--- a/src/lib/teads-curve/index.ts
+++ b/src/lib/teads-curve/index.ts
@@ -2,6 +2,11 @@ import {IImpactModelInterface} from '../interfaces';
 import Spline from 'typescript-cubic-spline';
 import {KeyValuePair} from '../../types/boavizta';
 
+export enum Interpolation {
+  SPLINE = 'spline',
+  LINEAR = 'linear',
+}
+
 export class TeadsCurveModel implements IImpactModelInterface {
   // Defined for compatibility. Not used in TEADS.
   authParams: object | undefined;
@@ -15,6 +20,8 @@ export class TeadsCurveModel implements IImpactModelInterface {
   points: number[] = [0, 10, 50, 100];
   // spline interpolation of the power curve
   spline: Spline = new Spline(this.points, this.curve);
+  // interpolation method
+  interpolation: Interpolation = Interpolation.SPLINE;
 
   /**
    * Defined for compatibility. Not used in TEADS.
@@ -53,6 +60,10 @@ export class TeadsCurveModel implements IImpactModelInterface {
       this.spline = new Spline(this.points, this.curve);
     }
 
+    if ('interpolation' in staticParams) {
+      this.interpolation = staticParams?.interpolation as Interpolation;
+    }
+
     return this;
   }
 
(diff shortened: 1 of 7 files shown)


src/lib/teads-curve/index.ts
```diff
@@ -0,0 +1,134 @@
...
+
+export class TeadsCurveModel implements IImpactModelInterface {
+  // Defined for compatibility. Not used in TEADS.
+  authParams: object | undefined;
+  // name of the data source
+  name: string | undefined;
+  // tdp of the chip being measured
+  tdp = 100;
+  // default power curve provided by the Teads Team
+  curve: number[] = [0.12, 0.32, 0.75, 1.02];
```
narekhovhannisyan, line 13: let's keep such configurations in config file

src/lib/teads-curve/index.ts
```diff
@@ -0,0 +1,134 @@
...
+  async calculate(
+    observations: object | object[] | undefined
+  ): Promise<object> {
+    if (observations === undefined) {
+      throw new Error('Required Parameters not provided');
+    }
+
+    const results: KeyValuePair[] = [];
+    if (Array.isArray(observations)) {
+      observations.forEach((observation: KeyValuePair) => {
```
narekhovhannisyan, line 76: you don't like reduce? :D

gnanakeethan: I am trying to keep it closer to Javascript rather than Typescript.

src/lib/teads-curve/index.ts
narekhovhannisyan: it's convenient when f.ex. you want to tune boavizta params, you go directly to config and you know that everything should be tuned from there instead of diving into codebase.

src/lib/teads-curve/index.ts
narekhovhannisyan: it's javascript-specific, and you don't have to declare results variable.

src/lib/teads-curve/index.ts
gnanakeethan: These plugins will be spinoff into a separate project sooner.  

The main point is that this configuration is adjustable while using the Impact Plugin. 

It will be unnecessary to extract them to a separate file, as these are **not constants** and are modified while using the plugin. 

These should be documented into the documentation rather than being extracted to a separate file for readability.

narekhovhannisyan: Got your point. So these values are dynamic rather than static.

src/lib/teads-curve/index.ts
narekhovhannisyan:
```
if (Array.isArray(observations)) {
  return observations.reduce((acc, observation) => {
    const energy = this.calculateEnergy(observation);
    acc.push({
      ...observation,
      energy,
    })
    
    return acc
  }, [])
}

```

narekhovhannisyan: this solution is more like separate component and is less coupled with parent function

src/lib/teads-curve/index.ts
jawache: In this case these are a magic set of numbers that are called the TEADs curve. So they won't ever change. Happy for them to remain as is.

src/lib/teads-curve/index.ts
```diff
@@ -0,0 +1,134 @@
...
+    name: string,
+    staticParams: object | undefined = undefined
+  ): Promise<IImpactModelInterface> {
+    this.name = name;
+
+    if (staticParams === undefined) {
+      throw new Error('Required Parameters not provided');
+    }
+
+    if ('tdp' in staticParams) {
```
jawache, line 42: tdp will come from static params or from the observation itself.

src/lib/teads-curve/index.ts
```diff
@@ -0,0 +1,134 @@
...
+      this.tdp = staticParams?.tdp as number;
+    } else {
+      throw new Error(
+        '`tdp` Thermal Design Power not provided. Can not compute energy.'
+      );
+    }
+
+    if ('curve' in staticParams) {
+      this.curve = staticParams?.curve as number[];
+      this.spline = new Spline(this.points, this.curve);
```
jawache, line 52: Similar point to the one I made on TEADs AWS. Can we have this configurable
interpolation: linear | curve (default is linear since that's how everyone is using this these days).
Also are you making the curve itself configurable?

src/lib/teads-curve/index.ts
jawache: Or so `acc` becomes a copy of the `observation` but with `energy` added in?

jawache (changes requested):
Thanks made some comments but happy to merge after those have been dealt with.

src/lib/teads-curve/index.ts
gnanakeethan: @jawache I intend to allow anyone to adjust the spline. Just keeping an option open at one end so plugin users can adjust instead of creating their own version of the implementation.

src/lib/teads-curve/index.ts
gnanakeethan: I feel like it is increasing cognitive complexity in a library code.

src/lib/teads-curve/index.ts
gnanakeethan: I am making the curve itself configurable. So giving in a points array of [0,100] and curve array of [0.1,1.05] for example in static params will make the computation in a linear manner. 

Linear interpolation can only performed between two points. I might need to do another approximation if we want to do multiple points and it is certainly not the appropriate one.

src/lib/teads-curve/index.ts
gnanakeethan: The result object contains the energy calculation + observation as given in the issue description.

narekhovhannisyan: @jawache @gnanakeethan 

```js
if (Array.isArray(observations)) {
  return observations.map((observation) => {
    observation.energy = this.calculateEnergy(observation)
    
    return observation
  })
}
```
doesn't use any array push, or spread operator.

gnanakeethan: I am using map. Reduce is something different.

gnanakeethan:
@jawache I have updated the documentation + code. The curve is no longer customizable.

jawache (approved):

PR #111, opened 2023-09-01
Rimpl cli script

narekhovhannisyan:
### Types of changes

- Rimpl calculation functionality.
-


src/util/observatory.ts
```diff
@@ -0,0 +1,63 @@
+import {BoaviztaCpuImpactModel} from '../lib';
+
+/**
+ * Pipeline for computing impacts based on observation.
+ */
+export class Observatory {
```
jawache, line 6: I like the name :)

src/util/observatory.ts
```diff
@@ -0,0 +1,63 @@
...
+   * Init observations object.
+   */
+  constructor(observations: any) {
+    this.observations = observations;
+  }
+
+  /**
+   * Apply `boavizta` model computation to observations.
+   */
+  private async boaviztaHandler(params: any) {
```
jawache, line 19: I know this is draft but eventually there should be a separate modelFactory which is initialised when the graph is loaded and during computation like this returns an instance of a model given the name that you are requesting.

src/util/observatory.ts
```diff
@@ -0,0 +1,63 @@
...
+    const modelInstance = await new BoaviztaCpuImpactModel().configure(
+      'test',
+      params
+    );
+
+    const preparedObservations = this.observations.map((observation: any) => ({
```
jawache, line 25: There does need to be some manipulation of data in a general sense but we should not have to hard code specific transformations for specific models. The data is either passed in a way that the model can take without any transformation, or if some transformation is required the model is responsible for doing it itself.

src/util/observatory.ts
```diff
@@ -0,0 +1,63 @@
...
+    const calculatedImpacts = await modelInstance.calculate(
+      preparedObservations
+    );
+
+    const result = this.observations.map((observation: any, index: number) => ({
+      ...observation,
+      ...calculatedImpacts[index],
+    }));
+
+    this.observations = result;
```
jawache, line 40: Store the ouput in a thing called impacts, yes it will cotain the same data as the observations but any impacts added in.

impacts:
...
obsevations:
...

jawache (commented):
Just adding comments, I know we're in draft.

src/util/observatory.ts
narekhovhannisyan: Agree, I have got rid of specific handlers locally (not pushed yet), trying to have a unified interface for all models.

src/util/observatory.ts
narekhovhannisyan: Agree, I don't like this either. It's better to tune yaml/model to use the same interface rather than mapping to one interface, then mapping the model result back to original pattern

scripts/rimpl.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    // Lifecycle Initialize Models
+    const modelsHandbook = new ModelsUniverse();
+
+    impl.initialize.models.forEach((model: any) =>
+      modelsHandbook.writeDown(model)
+    );
+
+    // Initialize impact graph/computing
+    const services = Object.keys(impl.graph.children);
```
jawache, line 55: @narekhovhannisyan FYI for PoC this is fine for now but it eventually needs to be something that parses the children, the children of children, etc.... and collects all the component nodes (there are grouping nodes and component nodes, component nodes are the nodes that hold observations). Component nodes will never have children.

Then we loop through the component nodes.

The name of the list of things should be "components" or "componentNodes" I would not call these services.

scripts/rimpl.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    for (const model of pipeline) {
+      const instance: any = await modelsHandbook.initalizedModels[model](
+        config[model]
+      );
+
+      await observatory.doInvestigationsWith(instance);
+    }
+
+    const impact = observatory.getObservedImpact();
```
jawache, line 24: Can we rename these functions to this? (to map the language used in the spec)

```suggestion
    const impacts = observatory.compute();
```

scripts/rimpl.ts
```diff
@@ -0,0 +1,72 @@
...
+ * For each graph builds params, then passes it to computing fn.
+ */
+const calculateImpactsBasedOnGraph =
+  (impl: any, modelsHandbook: ModelsUniverse) => async (service: string) => {
+    const serviceData = impl.graph.children[service];
+    const {pipeline, observations, config} = serviceData;
+
+    const observatory = new Observatory(observations);
+
+    for (const model of pipeline) {
```
jawache, line 16:
```suggestion
    for (const modelName of pipeline) {
```

scripts/rimpl.ts
```diff
@@ -0,0 +1,72 @@
...
+      const instance: any = await modelsHandbook.initalizedModels[model](
```
jawache, line 17: Took me a while to figure out what the variables names meant, just suggesting alternatives that are more descriptive

```suggestion
      const modelInstance: any = await modelsHandbook.initalizedModels[modelName](
```

src/util/models-universe.ts
```diff
@@ -0,0 +1,102 @@
...
+   * Returns shell model.
+   */
+  private handShellModel() {
+    return ShellModel;
+  }
+
+  /**
+   * Gets model based on `kind` and `name` params.
+   */
+  private handModelByCriteria(kind: string, name: string) {
```
jawache, line 65: If you want to handle plugin and shell you will also need to pass in path, since plugin and shell will need path to dynamically load.

jawache (commented):
I might make more suggestions in terms of naming things in the future and not had a chance to run it locally yet but this looks like it works as expected from looking at the code.

scripts/rimpl.ts
narekhovhannisyan: yes we may need recursive approach, this flat version just for having working structure

src/util/models-universe.ts
narekhovhannisyan: haven't handled that cases yet

PR #169, opened 2023-09-20
Update sci model

### LINKED ISSUE #155: Model Plugin - sci
- **What**: An impact model that calculates an SCI score
- **Why**: So we can compute an SCI score with a functional unit.
- **DoD**:
- [ ] Does the model meet the specification?
- [ ] Does the model have complete test coverage, including edge cases, happy path and error cases?
- [ ] Is the code documented?

## Config
- `functional-unit-duration`: Default is none, if it's provided then we are calculating an SCI score where the functional unit is a time span, e.g. carbon per hour. For details on the implementation head to the Behavior section below.
- 
## Observation
_List the variables this model expects in an input observation_

- `carbon`: The amount of carbon calculated for this observation.
- `functional-unit`: The functional unit denominator for an SCI calculation, the R in (`C per R`). It can be a numerical value or it can point to another field in the observation.

## Impacts
_List the variables this model exports out as an impact_
- `sci`: The sci value as defined by this manifest file, e.g. carbon per hour, or carbon per user.

## Behavior
_What does the model do to turn the observations into impacts? List any important equations or references._

Divides the `carbon` by `functional-units` and outputs the result as `sci`.
For this model to work the carbon already needs to be calculated but also the functional-units needs to be provided by the user.
If the functional unit is _users_ then we would expect the observation to contain either a field called "users" and the functional-unit field to point to that field like so:

```yaml
observations: 
  - timestamp: 2023-07-06T00:00
    duration: 3600
    users: 100
    functional-unit: users
```

Or the functional-unit itself can contain the numerical value like so:

```yaml
observations: 
  - timestamp: 2023-07-06T00:00
    duration: 3600
    functional-unit: 100
```

If the functional unit is for a time period then we need to do something special. 
The carbon value for an observation is assumed to be for the duration of that observation. 
If the duration is for 1 minute but the functional-unit is per hour, so every minute we want to report the carbon as if it's carbon per hour, then we will need to multiply the carbon value by 60 (or divide by 1/60). 
To handle this case we include the **global** configuration of `functional-unit-duration`, if this is present in the global config then:
- `functional-unit` is not required in the observation.
- `carbon` will be divided by `observation:duration`/`config:functional-unit-duration`

```yaml
config:
  functional-unit-duration: 216000 # functional unit of 1 hour in seconds.
observations: 
  - timestamp: 2023-07-06T00:00
    duration: 3600
```

## Assumptions
_What are the core assumptions this model makes?_
- You are calculating an SCI score.

## Limitations
In what situations is this model limited in it's function, behavior, accuracy etc...
- N/A

## Example
Provide at least one example of some input config and observations. This is in pimpl format not rimpl format, so the config is the global config and all parameters must be present on the observation.

### Input

```yaml
config:
  functional-unit-duration: 216000
observations: 
  - timestamp: 2023-07-06T00:00
    duration: 3600
    carbon: 100
```

### Output

```yaml
config:
  functional-unit-duration: 216000
observations: 
  - timestamp: 2023-07-06T00:00
    duration: 3600
    carbon: 100
impacts: 
  - timestamp: 2023-07-06T00:00
    duration: 3600
    carbon: 100
    sci: 6000 # 100 / (3600 / 21600)
```

jmcook1186:
Updates existing `sci` model implementation to match spec outlined in this issue: https://github.com/Green-Software-Foundation/ief/issues/155

Implements the following logic:

- `sci` model config takes `functional_unit_duration`  `functional_unit_time` and `functional_unit` args
- `functional_unit_time` is a string describing the time unit ('minute', 'hour', etc) - multiplies up sci values assuming they arrive in units of /s
- `functiona;_unit_duration` is the number of seconds an observation represents
- `functional_unit` is a string describing the functional unit, e.g. `requests`. There is then expected to be a `requests` field in each observation that is used to scale the sci value.
- if a `carbon` value is present, it is used. If not, then `carbon` is calculated as the sum of `operational_carbon` + `embodied-carbon`

A carbon value is then divided by `functional_time_duration` to convert to C/s. It is then multiplied up to be represented in units of C/`functional_unit_time` (e.g. C/day). It is then divided by `functional_unit` to give C/time/unit e.g. C/day/user

- Updates existing example impls to conform to this change

Fixes #155


gnanakeethan (approved):

narekhovhannisyan (approved):
lgtm, by the way please check contributing file :D

PR #204, opened 2023-10-13
docs: add shell model

jmcook1186:
### Types of changes

Adds documentation and example impl for the shell model

### DIFF
--- a/docs/implementations/shell.md
+++ b/docs/implementations/shell.md
@@ -1,0 +1,46 @@
+# Shell-imp
+
+The `shell-imp` is a wrapper enabling models external to IEF to be executed as part of an IEF pipeline. For example, you might have a standalone model written in Python. `shell-imp` spawns a subprocess to execute that Python model in a dedicated shell and pipes the results back into IEF's Typescript process.
+
+## Implementation
+
+To run the model, you must first create an instance of `ShellModel` and call its `configure()` method. The `configure` method takes `executable` as an argument - this is a path to an executable file. Then, you can call `calculate()` to run the external model.
+
+```typescript
+const impactModel = new ShellModel();
+await impactModel.configure('test', {
+    executable: '/usr/local/bin/sampler',
+});
+const result = await impactModel.calculate([
+    {duration: 3600, cpu: 0.5, datetime: '2021-01-01T00:00:00Z'}
+    ])
+```
+
+## Example impl
+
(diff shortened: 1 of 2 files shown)


docs/implementations/shell.md
```diff
@@ -0,0 +1,54 @@
...
+      config:
+        sampler:
+          executable: python3 /usr/local/bin/sampler
+      observations:
+        - timestamp: 2023-07-06T00:00
+          duration: 1 # Secs
+
+```
+
+In this example, the model invoked by executing `python3 /usr/local/bin/sampler` in a shell.
```
gnanakeethan, line 54: I think we need to give an example for the input for the shell model & the output from the shell binary itself

jmcook1186: done!

gnanakeethan: Can I make add more?

jmcook1186: yes, please do

narekhovhannisyan:
@jmcook1186 you can drop braces from commit messages since it takes the capacity of the message.

docs/implementations/shell.md
```diff
@@ -0,0 +1,87 @@
+# Shell-imp
+
+The `shell-imp` is a wrapper enabling models external to IEF to be executed as part of an IEF pipeline. For example, you might have a standalone model written in Python. `shell-imp` spawns a subprocess to execute that Python model in a dedicated shell and pipes the results back into IEF's Typescript process.
+
```
jmcook1186, line 4:
```suggestion

## Parameters
### Model config
The model should be intialized with a name and `kind: shell`, as follows:

The shell model interface requires a path to the model executable. This path is provided in the model configuration with the name `executable`. The path should be appended by the execution command, for example if the executable is a binary, the path would be prepended with `./` on a Linux system. If the model is intended to be run as Python, you can prepend `python`.

- `executable`: the path to the model executable along with the execution command as it would be entered into a shell.

### inputs
The parameters included in the `inputs` field int he `impl` depend entirely on the model itself. A typical model plugin might expect the following common data to be provided as `inputs`: 
- `timestamp`: A timestamp for the specific input observation
- `duration`: The length of time these specific inputs cover
- `cpu-util`: The percentage CPU utilization for the specific duration.

## Returns

The specific return types depend on the model being invoked. Typically, we would expect some kind of energy or carbon metric as an output, but it is also possible that models target different parts of the pipeline, such as data importers, adaptor models etc. Therefore, we do not specify return data for external models.

```

docs/implementations/shell.md
jmcook1186: no action on this - please raise a separate PR to add your examples later

gnanakeethan (approved):

PR #208, opened 2023-10-17
[models]: update use of `duration` and `functional-unit-duration` in `sci` model

### LINKED ISSUE #207: Update how SCI handles `duration`
Right now we use the `duration` ans `functional-unit-duration` in an overcomplicated way with unnecessary duplication. We can make a relatively simple change to a) simplify the logic, and b) add some functionality.

## What happens now

`sci` takes three parameters: `functional-unit`, `functional-unit-time` and `functional-unit-duration`. 
`functional-unit-duration` is used to force the `carbon` measurement to be in units of /s at the start of the calculation, enabling conversions to other time units (/d, /m, etc). This means the `functional-unit-duration` is the same as the `duration` already provided in each observation. 

`functional-unit-duration` being equal to `duration` also means that we are limited to whole units for time conversions (e.g. you can only provide a SCI value in time units of whole s, m, d, y and never, say, half a day, or 2 months.

## What should happen 

We should force the incoming `carbon` value into units of /s using `duration` from each observation. `functional-unit-duration` should be the number of `functional-unit-time` to normalize `sci` to. The flow would then be:

```
carbon / duration --> carbon/s
carbon/s --> time conversion function --> carbon/whole-unit-of-time
carbon/whole-unit-of-time * functional-time-duration --> c/desired-unit-of-time
```

We would normally expect `functional-unit-duration` to be `1`, as most people will want their `sci` values in /m, /d, /month, /yr etc. But as well as simplifying the calculation logic, this way allows people to predict `sci` for *any* time value.

Here's an arbitrary example with hypothetical, not necessarily sensible, numbers:

`carbon = 20`
`functional-unit = requests`
`functional-unit-time = d`
`functional-unit-duration = 2`
`requests = 1000`
`duration = 2`

the calculation would be:

```
carbon/duration = carbon/s = 20/2 = 10
carbon/s * 60 * 60 * 24 = carbon/d = 864000
carbon/d * `functional_unit_duration` = carbon/d * 2d = 1728000
carbon/2d / requests = carbon/requests/2d = 1728.0
```

jmcook1186:
### Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

Updates `sci` model to use `duration` from each observation to ensure initial carbon value is expressed in seconds, then uses `functional-unit-duration` to determine the time span for the final `sci` estimate. See Issue #207 for detailed explanation of the problem.

Also updates docs to reflect changes. 

Fixes #207

### DIFF
--- a/src/lib/sci/index.ts
+++ b/src/lib/sci/index.ts
@@ -1,11 +1,11 @@
-import {IImpactModelInterface} from '../interfaces';
+import { IImpactModelInterface } from '../interfaces';
 
-import {CONFIG} from '../../config';
+import { CONFIG } from '../../config';
 
-import {KeyValuePair} from '../../types/common';
+import { KeyValuePair } from '../../types/common';
 
-const {MODEL_IDS} = CONFIG;
-const {SCI} = MODEL_IDS;
+const { MODEL_IDS } = CONFIG;
+const { SCI } = MODEL_IDS;
 
 export class SciModel implements IImpactModelInterface {
   authParams: object | undefined = undefined;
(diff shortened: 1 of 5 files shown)


narekhovhannisyan (approved):

src/lib/sci/index.ts
```diff
@@ -89,14 +98,20 @@ export class SciModel implements IImpactModelInterface {
         sci_timed = sci_secs * 60 * 60 * 24 * 365;
       }
 
+      /*
+      sci currently in whole single units of time - multiply by duration to
+      convert to user-defined span of time.
+      */
```
narekhovhannisyan, line 104: /*
 *
 */

narekhovhannisyan: you can auto generate it by writing `/**` and pressing enter

jawache (approved):

PR #351, opened 2023-12-20
Time sync

### LINKED ISSUE #170: Model Plugin - time-sync
- **What**: A core "built-in" model plugin which synchronizes all the observations in the provided series into a series.
- **Why**: So we can place all the outputs and impacts onto a grid which can then be aggregated up.
- **DoD**:
- [ ] Does the model meet the specification?
- [ ] Does the model have complete test coverage, including edge cases, happy path and error cases?
- [ ] Has documentation been written for someone writing an IMPL using this model?
- [ ] Has the `units.yml` been updated to take into consideration the params used in this model

## Config
- `start-time`: The date-time to start the series.
- `end-time`: The end time to finish the series.
- `duration`: The number of seconds in each interval duration.
- `count`: The number of buckets in the series.
- `weighting`: The algorithm to use to weight the values in buckets, the default is `time` (the only current algorithm is time)

**NOTE**: Both start-time and end time cannot be provided at the same time, it's either one or the other.

## Observation
_List the variables this model expects in an input observation_
- N/A

## Impacts
_List the variables this model exports out as an impact_
- N/A

### Overview

- Every component has a time series of observations but there is no guarantee they match is duration, times, overlap. No guarantee of any form at all.
- This model define a global time series, with a fixed start and end time, consistent set of buckets of the same duration.
- The goal of this model is to snap all the time series for all the components (where this model is configured on the pipeline) to that global time series.

### Global Time Series

If `start-time` has been defined then we create a series which starts at the start time and then moves forward in time. Each bucket `duration` seconds long and there are `count` buckets. We start in the past and move forward in time.

If `end-time` has been defined then we create a time series which **ends** at end time and then create the time series by going backwards in time.

We suspect the most useful use case is to choose an end-time of now and go backwards rather than to choose a start time and go forwards but we provide both.

**NOTE**: A possible v2 feature might be a way for models to scan the whole graph and automatically determine the start-end times from the observations present in the system.

### Normalization

The goal of this model is to snap the observations to this global time series.

>[!note]
> The output observation might be zero for a given time and duration. It means nothing was running then, so there could not be any emissions from the component.

An observation might be of a longer or smaller duration than the new globally defined duration or overlaps the new duration in some way. There can be many algorithms we can use to bucket/slice up the values, but a good default strategy is to use a weighting of time. 

### Example

Take, for example the below, it has mixed durations, some are 5, and some are 15mins.

| Timestamp        | Duration | Energy | Location | Grid | Operational |
| ---------------- | -------- | ------ | -------- | ---- | ----------- |
| 2023-07-06T00:00 | 15       | 43     | west-us  | 500  | 4.7         |
| 2023-07-06T15:00 | 5        | 20     | west-us  | 490  | 2.9         |
| 2023-07-06T20:00 | 5        | 18     | west-us  | 470  | 2.8         |

If we had decided to bucket the outputs in 5 mins increments, we would specify a `duration` of 5 mins. How to normalize the buckets to 5 min durations?

This effectively boils down to a choice of what weighting to use. The simplest would be to weight by time. So the first duration is 15 mins long; if we wanted to visualize in 5 min buckets, we would divide the impacts by 3.

| Timestamp        | Duration | Energy | Location | Grid | Operational |
| ---------------- | -------- | ------ | -------- | ---- | ----------- |
| 2023-07-06T00:00 | 5        | 14.3   | west-us  | 500  | 1.57        |
| 2023-07-06T05:00 | 5        | 14.3   | west-us  | 500  | 1.57        |
| 2023-07-06T10:00 | 5        | 14.3   | west-us  | 500  | 1.57        |
| 2023-07-06T15:00 | 5        | 20     | west-us  | 490  | 2.9         |
| 2023-07-06T25:00 | 5        | 18     | west-us  | 470  | 2.8         |

>[!important]
> The decision to **weight** an observation to the new duration or just copy it is determined by its aggregation param in the units.yaml https://github.com/Green-Software-Foundation/ief/blob/dev/src/config/units.yaml. If the aggregation is "sum" then we would divide it. If the aggregation is avg (or another) then we simply copy it. **This needs further thought and clarification, there may be many other cases that we are not thinking about clearly here.**

## Assumptions
What are the core assumptions this model makes?
- One core assumption is that a time weighting is the right approach.
- Another assumption is that we can decide the weighting using the aggregation param in the units.yaml.

## Limitations
In what situations is this model limited in it's function, behaviour, accuracy etc...

## Example
Provide at least one example of some input config and observations. This is in pimpl format not rimpl format, so the config is the global config and all parameters must be present on the observation.

### Input

```yaml
config:
  start-time: 2023-07-06T00:00
  duration: 300
  count: 5
  weighting: time
observations: 
  - timestamp: 2023-07-06T00:00
    duration: 900
    location: west-us
    energy: 42.9
    grid-ci: 500
    operational-carbon: 4.71    
  - timestamp: 2023-07-06T15:00
    duration: 300
    location: west-us
    energy: 20
    grid-ci: 490
    operational-carbon: 2.9
  - timestamp: 2023-07-06T25:00
    duration: 300
    location: west-us
    energy: 18
    grid-ci: 470
    operational-carbon: 2.8       
```

### Output

```yaml
config:
  start-time: 2023-07-06T00:00
  duration: 300
  count: 5
  weighting: time
impacts:
  - timestamp: 2023-07-06T00:00
    duration: 300 
    location: west-us
    energy: 14.3
    grid-ci: 500
    operational-carbon: 1.57
  - timestamp: 2023-07-06T05:00
    duration: 300
    location: west-us
    energy: 14.3
    grid-ci: 500
    operational-carbon: 1.57
  - timestamp: 2023-07-06T10:00
    duration: 300
    location: west-us
    energy: 14.3
    grid-ci: 500
    operational-carbon: 1.57    
  - timestamp: 2023-07-06T15:00
    duration: 300
    location: west-us
    energy: 20
    grid-ci: 490
    operational-carbon: 2.9
  - timestamp: 2023-07-06T25:00
    duration: 300
    location: west-us
    energy: 18
    grid-ci: 470
    operational-carbon: 2.8
observations: 
  - timestamp: 2023-07-06T00:00
    duration: 900
    location: west-us
    energy: 42.9
    grid-ci: 500
    operational-carbon: 4.71    
  - timestamp: 2023-07-06T15:00
    duration: 300
    location: west-us
    energy: 20
    grid-ci: 490
    operational-carbon: 2.9
  - timestamp: 2023-07-06T25:00
    duration: 300
    location: west-us
    energy: 18
    grid-ci: 470
    operational-carbon: 2.8    
```

### LINKED ISSUE #286: Add normalization model to `if-models`
## Story
As a user, I want to be able to automatically normalize my outputs to a standard temporal resolution so that I can generate time series and more accurately see trends in my impact over time. 

## Rationale

Each individual observation has its own specified `duration` which can differ from observation to observation. This prevents users from being able to easily export time series of their impacts and aggregate their results appropriately, because time could be erroneously squashed or stretched by algorithms assuming each observation to represent equal durations. With normalization, outputs are equally spaced along a uniform timeline.

Note, that we will **only have duration and total impact** data available, which yields the following valid high level approaches:
1) upscale by extrapolation (then we discuss which extrapolation model)
2) upscale assuming no additional expenditure (representing the observation capturing a burst and then no usage)
3) downscale by spreading existing data across more time intervals (then we discuss what model to use to spread the impact)
4) downscale by assuming some non-uniform distribution function

There should be weighting options, and the ability to upscale and downscale.

**upscaling**
- `extrapolate-linear`: to upscale from `duration` to `norm_duration`, assume the measured impact will continue at the same rate across the longer time period, e.g. 3 gC/s becomes 180 gC/min
- `equal-spread`: upscale to longer `norm_duration` by 'spreading' the measured impact across the new timespan (i.e. total impact is not changes, but it is distributed over a longer time), e.g. 3gC/s becomes 3gC/min 

**downscaling**

- `equal-spread`: assume the total impact is equally distributed across the smaller `durations`, i.e. 3gC/min becomes 60 observations each reporting 0.05 gC/s
- `distribute`: assume the impact is distributed over the measured `duration`, for example normal distribution, some known curve or skew. **This needs to be defined before this option can be implemented**

**Q** What should we do if upscaling time leads to overlapping timestamps? Reject? Apply some correction? 

## Implementation details
- Create a model, `NormModel`
- The model will operate on each observation independently.
- All times are provided in units of seconds
- The model logic will be as follows:
    - check the `norm_duration` and `weighting_method` that will be available in each observation
    - if the `norm_duration` > `duration` then enter upscaling branch
       - if method is `extrapolate_linear` in upscaling branch determine `multiplier` required to convert `duration` to 
       -   `norm_duration` (e.g. for 1s -> 1 min it is 60). Then return one observation with `duration` == `norm_duration` and each impact overwritten to `impact * multiplier`.
        - if method is `equal-spread` return one observation with `duration` == `norm-duration` and each impact is left unchanged.
    - if `norm_duration` < `duration` then enter the downscaling branch
    - if method is `equal_spread` calculate denominator as `norm_duration/duration` (e.g. for mins -> seconds, 60/1 = 60)
        - return `floor(denominator)` number of observations, each with `impact/denominator` values
    - if method is `distribute` apply the chosen distribution function, amintaining the same total impact over the new timespan.

Note that this will mean the number of observations in `inputs` will no longer be the same as the number of observations in `outputs`.

## Priority
3/5 this is a necessary feature that should be shipped asap, but there's no specific external deadline .

## Scope

As this will be implemented as a model plugin, it should be purely additive and should not cause any breakages to other parts of the framework.

## Size
Estimate 2 days for implementation, testing and documentation updates

## What does "done" look like?
- model exists in `if-models` repository
- model implements all logic outlined above
- unit tests exist that provide 100% coverage for the new model
- unit tests pass
- at least one demo impl has been created
- the demo impl runs as expected
- documentation has been added to the docs website
- documentation has been added to the model README in the `if-models` repo

## Does this require updates to documentation or other materials??

Yes, documentation should be added to the model README in `if-models` repo, and implementation details should eb added to the documentation website.

## What testing is required?
- Unit tests with 100% coverage
- demo impl that runs as expected

## Is this a known/expected update?

This has been a feature on the roadmap since inception, and has been included in the project specification.
Implementation details will be discussed with team before this issue is made READY.

narekhovhannisyan:
### Types of changes

- Introduced time sync builtin model.

closes #170 
closes #286

### DIFF
--- a/src/models/time-sync.ts
+++ b/src/models/time-sync.ts
@@ -79,9 +79,9 @@ export class TimeSyncModel implements ModelPluginInterface {
   }
 
   /**
-   * Populates object to fill the gaps in observation timeline.
+   * Populates object to fill the gaps in observational timeline using zeroish values.
    */
-  private inputFiller(
+  private fillWithZeroishInput(
     input: ModelParams,
     missingTimestamp: number,
     dealer: UnitsDealerUsage
@@ -110,10 +110,7 @@ export class TimeSyncModel implements ModelPluginInterface {
       }
 
       const method = dealer.askToGiveMethodFor(metric);
-      acc[metric] =
-        method === 'sum'
-          ? this.convertPerInterval(input[metric], input['duration'])
-          : input[metric];
+      acc[method] = method === 'avg' || method === 'sum' ? 0 : input[metric];
 
       return acc;
     }, {} as ModelParams);
@@ -170,7 +167,7 @@ export class TimeSyncModel implements ModelPluginInterface {
               missingTimestamp <= currentMoment.valueOf() - 1000;
               missingTimestamp += 1000
             ) {
-              const filledGap = this.inputFiller(
+              const filledGap = this.fillWithZeroishInput(
                 input,
                 missingTimestamp,
                 dealer
--- a/src/models/time-sync.ts
+++ b/src/models/time-sync.ts
@@ -1,18 +1,18 @@
 import moment = require('moment');
 
-import { ERRORS } from '../util/errors';
-import { STRINGS } from '../config';
+import {ERRORS} from '../util/errors';
+import {STRINGS} from '../config';
 
-import { UnitsDealer } from '../util/units-dealer';
+import {UnitsDealer} from '../util/units-dealer';
 
-import { ModelParams, ModelPluginInterface } from '../types/model-interface';
-import { TimeNormalizerConfig } from '../types/time-sync';
-import { UnitsDealerUsage } from '../types/units-dealer';
-import { UnitKeyName } from '../types/units';
+import {ModelParams, ModelPluginInterface} from '../types/model-interface';
+import {TimeNormalizerConfig} from '../types/time-sync';
+import {UnitsDealerUsage} from '../types/units-dealer';
+import {UnitKeyName} from '../types/units';
 
-const { InputValidationError } = ERRORS;
+const {InputValidationError} = ERRORS;
 
-const { INVALID_TIME_NORMALIZATION, INVALID_TIME_INTERVAL } = STRINGS;
+const {INVALID_TIME_NORMALIZATION, INVALID_TIME_INTERVAL} = STRINGS;
 
 export class TimeSyncModel implements ModelPluginInterface {
   startTime: string | undefined;
--- a/src/models/time-sync.ts
+++ b/src/models/time-sync.ts
@@ -142,7 +142,7 @@ export class TimeSyncModel implements ModelPluginInterface {
   /**
    * Checks if padding is needed either at start of the timeline or the end and returns status.
    */
-  private checkPadding(inputs: ModelParams[]): PaddingReceipt {
+  private checkPad(inputs: ModelParams[]): PaddingReceipt {
     const startDiffInSeconds =
       moment(inputs[0].timestamp).diff(moment(this.startTime)) / 1000;
 
@@ -200,6 +200,22 @@ export class TimeSyncModel implements ModelPluginInterface {
     return paddedArray;
   }
 
+  private trimInputs(inputs: ModelParams[]): ModelParams[] {
+    const trimmedInputs: ModelParams[] = [];
+
+    inputs.forEach(input => {
+      console.log(input);
+      if (
+        moment(input.timestamp).isSameOrAfter(moment(this.startTime)) &&
+        moment(input.timestamp).isSameOrBefore(moment(this.endTime))
+      ) {
+        trimmedInputs.push(input);
+      }
+    });
+    // console.log("trimmedInputs: ", trimmedInputs)
+    return trimmedInputs;
+  }
+
   /**
    * Normalizes provided time window according to time configuration.
    */
@@ -207,7 +223,7 @@ export class TimeSyncModel implements ModelPluginInterface {
     this.validateParams();
 
     const dealer = await UnitsDealer();
-    const pad = this.checkPadding(inputs);
+    const pad = this.checkPad(inputs);
     const paddedInputs = this.padInputs(inputs, pad, dealer);
 
     return paddedInputs
@@ -265,7 +281,9 @@ export class TimeSyncModel implements ModelPluginInterface {
           acc.push(normalizedInput);
         }
 
-        return acc;
+        const trimmedInputs = this.trimInputs(acc);
+
+        return trimmedInputs;
       }, [] as ModelParams[])
       .sort((a, b) => moment(a.timestamp).diff(moment(b.timestamp)));
   }
--- a/src/models/time-sync.ts
+++ b/src/models/time-sync.ts
@@ -3,7 +3,7 @@ import {ERRORS} from '../util/errors';
 import {STRINGS} from '../config';
 
 import {ModelParams, ModelPluginInterface} from '../types/model-interface';
-import {TimeNormalizerConfig} from '../types/time-normalization';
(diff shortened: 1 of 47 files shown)


jmcook1186 (approved):
There was a mistake with the value returned from units.yaml. With that fixed, this appears to be working well - I tested with several different "broken" time series and they were all processed correctly.

src/models/time-sync.ts
```diff
@@ -224,64 +276,63 @@ export class TimeSyncModel implements ModelPluginInterface {
...
+      const currentMoment = moment(input.timestamp);
+
+      /** Checks if not the first input, then check consistency with previous ones. */
+      if (index > 0) {
+        const previousInput = paddedInputs[index - 1];
+        const previousInputTimestamp = moment(previousInput.timestamp);
+
+        /** Checks for timestamps overlap. */
+        if (
+          previousInputTimestamp
```
jmcook1186, line 292: `previousInputTimestamp.add()` updates the value of `previousTimestamp` in-place, breaking the cross-observation in-filling.

jmcook1186: replace with new `moment` instance instead to fix

jmcook1186 (approved):
There was a bug relating to the in-place updating of a timestamp inside a conditional statement that was causing the cross-observation infilling to break. This is now fixed. I also added some spelling corrections and added some brief comments where I thought they were helpful. Tested manually with a range of unusual time conditions and now seems to be working well. Merging.

PR #387, opened 2024-01-12
examples: in impls and ompls update shell.yml files

manushak:
### Types of changes

- made changes in shell.yml files in impls and ompls test folders
-

### DIFF
--- a/examples/impls/test/shell.yml
+++ b/examples/impls/test/shell.yml
@@ -1,5 +1,5 @@
-name: ccf-demo
-description: example impl invoking CCF model
+name: shell-demo
+description: example impl invoking shell model
 tags:
 initialize:
   models:
--- a/examples/impls/test/shell.yml
+++ b/examples/impls/test/shell.yml
@@ -11,7 +11,9 @@ graph:
     child:
       pipeline:
         - sampler
+      config:
+        sampler:
+          command: python3 /usr/local/bin/sampler
       inputs:
         - timestamp: "2023-11-02T10:35:31.820Z"
           duration: 3600
-          command: python3 /usr/local/bin/sampler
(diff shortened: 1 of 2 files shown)


examples/impls/test/shell.yml
```diff
@@ -11,9 +11,7 @@ graph:
...
-      config:
-        sampler:
-          executable: python3 /usr/local/bin/sampler
       inputs:
-          - timestamp: '2023-11-02T10:35:31.820Z'
-            duration: 3600
\ No newline at end of file
+        - timestamp: "2023-11-02T10:35:31.820Z"
+          duration: 3600
+          command: python3 /usr/local/bin/sampler
```
jmcook1186, line 17: `command` should be provided in the model config above, rather than in an individual `input`
e.g.

```
child:
  pipeline:
    - sampler
  config:
    sampler:
      command: python3 /usr/local/bin/sampler
  inputs:
    - timestep: "2023-11-02T10:35:31:820Z"
      energy: 1
      duration: 10
```

p.s. your version will run correctly *in this case* where there is only one timestep, but if there are multiple timesteps it will break unless you manually copy the command into each one. Putting it in the config means the value its autromaticaly copied to every `input`.

jmcook1186 (changes requested):
left a comment about the position of `command`

examples/impls/test/shell.yml
manushak: So, this PR doesn't have any sense?

jmcook1186: no, as long as `executable` is renamed to `command` that was the only required change to the impls

manushak: done

jmcook1186 (approved):
made some more small changes that I noticed, otherwise sgtm

PR #395, opened 2024-01-18
time-sync: added option to error out instead of padding

pazbardanl:
### Types of changes

- Added a flag to time-sync model that causes the model to error out instead of zero/zeroish padding (if needed)


examples/impls/test/time-sync-avoid-padding.yml
```diff
@@ -0,0 +1,50 @@
...
+              cpu-util: 10    
+              requests: 300
+            - timestamp: 2023-12-12T00:00:10.000Z
+              duration: 30
+              cpu-util: 20    
+              requests: 380
+            - timestamp: 2023-12-12T00:00:50.000Z
+              duration: 20
+              cpu-util: 20    
+              requests: 380
```
narekhovhannisyan, line 50: newline is missing at the end of file, please check

src/__tests__/unit/models/time-sync.test.ts
```diff
@@ -909,6 +923,7 @@ describe('execute(): ', () => {
       'start-time': '2023-12-12T00:00:00.000Z',
       'end-time': '2023-12-12T00:00:09.000Z',
       interval: 5,
+      'error-on-padding': false,
```
narekhovhannisyan, line 926: for fully covering the changes, it will be good to add one more test case where panic is enabled, and check if it's doing well

src/models/time-sync.ts
```diff
@@ -326,6 +336,9 @@ export class TimeSyncModel implements ModelPluginInterface {
 
         /** Checks if there is gap in timeline. */
         if (timelineGapSize > 1) {
+          if (this.errorOnPadding) {
+            throw new Error('avoiding padding on timeline gap');
```
narekhovhannisyan, line 340: please, init error messages in config/strings.ts file

src/models/time-sync.ts
```diff
@@ -143,6 +144,14 @@ export class TimeSyncModel implements ModelPluginInterface {
     }, {} as ModelParams);
   }
 
+  private errorOnPaddingIfNeeded(pad: PaddingReceipt): void {
+    if (this.errorOnPadding && (pad.start || pad.end)) {
+      const paddingDescription =
+        pad.start && pad.end ? 'start and end' : pad.start ? 'start' : 'end';
+      throw new Error('avoiding padding at ' + paddingDescription);
```
narekhovhannisyan, line 151: we have to use custom errors for being handled by framework

examples/impls/test/time-sync-avoid-padding.yml
pazbardanl: Done

src/__tests__/unit/models/time-sync.test.ts
pazbardanl: absolutely. Done.

src/models/time-sync.ts
```diff
@@ -143,6 +145,14 @@ export class TimeSyncModel implements ModelPluginInterface {
     }, {} as ModelParams);
   }
 
+  private validatePadding(pad: PaddingReceipt): void {
```
narekhovhannisyan, line 148: this method can be rewritten like this to be more organized, with doc

```
  /**
   * Checks if `error on padding` is enabled and and padding is needed. If so, then throws error.
   */
  private validatePadding(pad: PaddingReceipt): void {
    const {start, end} = pad;
    const isPaddingNeeded = start || end;

    if (this.errorOnPadding && isPaddingNeeded) {
      const paddingDescription =
        start && end ? 'start and end' : start ? 'start' : 'end'; // this padding description thing can be incapsulated in AVOIDING_PADDING logic. BTW function strings have unit tests

      throw new InputValidationError(AVOIDING_PADDING(paddingDescription));
    }
  }
```

src/models/time-sync.ts
```diff
@@ -19,13 +19,15 @@ const {
...
   INVALID_OBSERVATION_OVERLAP,
+  AVOIDING_PADDING,
 } = STRINGS;
 
 export class TimeSyncModel implements ModelPluginInterface {
   private startTime!: string;
   private endTime!: string;
   private dealer!: UnitsDealerUsage;
   private interval = 1;
+  errorOnPadding = false;
```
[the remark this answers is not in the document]
pazbardanl: yup, Done

src/models/time-sync.ts
pazbardanl: i like it, much cleaner code. Thanks :) also added tests for string functions. 
notice i put 2 different string functions since there is 1 where description is free text and the other is determined based on flags (start, end)

examples/impls/test/time-sync-avoid-padding.yml
```diff
@@ -0,0 +1,50 @@
...
+initialize:
+  models:
+    - name: time-synchronization
+      model: TimeSyncModel
+      path: builtin
+      config: 
+        start-time: 2023-12-12T00:00:00.000Z # ISO timestamp
+        end-time: 2023-12-12T00:01:00.000Z # ISO timestamp
+        interval: 5 # seconds
+        error-on-padding: true
```
jmcook1186, line 13: @narekhovhannisyan @pazbardanl minor point, but interested to hear your opinion on the field name. I'd suggest the name should be more positive action oriented, such as `allow-padding` and having `true` being the enabler and `false` triggering the exception, rather than `true` being the error case.
(not at all blocking)

pazbardanl: agree. it's more intuitive and requires less 'double-negative' to understand the behavior. ill modify accordingly.

src/models/time-sync.ts
```diff
@@ -20,14 +20,15 @@ const {
   INVALID_TIME_INTERVAL,
   INVALID_OBSERVATION_OVERLAP,
   AVOIDING_PADDING,
+  AVOIDING_PADDING_BY_EDGES: AVOIDING_PADDING_PER_RECEIPT,
```
narekhovhannisyan, line 23: I think we can introduce it with the same name in both places, renaming is unecessary

pazbardanl: true. fixed

narekhovhannisyan (approved):

Issue #411, opened 2024-01-25
CSV Exporting Functionality


jawache:
## Story
As a user, I want to export the entire graph as a CSV to analyze the data in other applications.

## Rationale
Trying to navigate and understand the impacts of your software application by looking at YAML is very challenging. By exporting into a flat table structure of a CSV, a human can better understand how the impacts are broken down by component and time.

## Implementation details

This is an extension to IF that outputs the tree in a CSV format.

ie --manifest <in> --exhaust csv --filter comma,seperated,list,of,parameters,to,export

The manifest is in the format of a YAML that has been processed by impact engine like so

```yaml

graph:
  outputs: # total per time bucket for the whole graph
    - timestamp: 2023-01-24T11:00
      duration: 360        
      carbon: 19
    - timestamp: 2023-01-24T11:05
      duration: 360        
      carbon: 12
    - timestamp: 2023-01-24T11:10
      duration: 360        
      carbon: 12
  aggregated:
    carbon: 43        
  children:
    application:
      outputs: # aggregated up the tree for every grouping node
        - timestamp: 2023-01-24T11:00
          duration: 360        
          carbon: 19
        - timestamp: 2023-01-24T11:05
          duration: 360        
          carbon: 12
        - timestamp: 2023-01-24T11:10
          duration: 360        
          carbon: 12      
      aggregated:
        carbon: 43          
      children:
        vm1:
          outputs: 
            - timestamp: 2023-01-24T11:00
              duration: 360        
              carbon: 3
            - timestamp: 2023-01-24T11:05
              duration: 360        
              carbon: 4
            - timestamp: 2023-01-24T11:10
              duration: 360        
              carbon: 5                  
          aggregated:
            carbon: 12
        vm2:
          outputs: 
            - timestamp: 2023-01-24T11:00
              duration: 360        
              carbon: 4
            - timestamp: 2023-01-24T11:05
              duration: 360        
              carbon: 5
            - timestamp: 2023-01-24T11:10
              duration: 360        
              carbon: 6    
          aggregated:
            carbon: 15                            
        vm3:
          outputs: 
            - timestamp: 2023-01-24T11:00
              duration: 360        
              carbon: 12
            - timestamp: 2023-01-24T11:05
              duration: 360        
              carbon: 3
            - timestamp: 2023-01-24T11:10
              duration: 360        
              carbon: 1                                                
          aggregated:
            carbon: 16                                        
```

The output CSV file should be of this format:

[...]

PR #437, opened 2024-02-13
Add aggregation feature

jmcook1186:
### Types of changes

Adds aggregation feature covering both vertical and horizontal aggregation.
Depending on the aggregation type defined by the user, the tree can be aggregated in vertical or horizontal mode, or both.

Vertical aggregation takes, for each metric defined in `aggregation.metrics`, the time series of observations for each child and aggregates them together, pushing the result to the parent node. If the parent node has its own data, it is aggregated too.

Horizontal aggregation takes each time series and sums the elements, giving a single representative number for the whole observation period.

If the aggregation method is `both` the following data are expected to be added to each parent node:

`total-{metric}`: the horizontally aggregated value for the parent node (not including children) (io.e. a number)
`total-aggregated-{metric}` - the horizontally aggregated value for the parent and all of its children (i.e. a number).
`aggregated-{metric}` - the vertically aggregated time series for the parent and all its children (i.e..a time series/ an array).

### DIFF
--- a/src/lib/aggregator.ts
+++ b/src/lib/aggregator.ts
@@ -1,58 +1,144 @@
 import {ERRORS} from '../util/errors';
 import {getAggregationMethod} from '../util/param-selectors';
-
+import {PARAMETERS} from '../config';
 import {STRINGS} from '../config';
 
-import {PluginParams} from '../types/interface';
-import {Parameters} from '../types/parameters';
-import {AggregationResult} from '../types/aggregation';
-
+import {ManifestCommon} from '../types/manifest';
+import {Node} from '../types/compute';
 const {InvalidAggregationParams} = ERRORS;
-const {INVALID_AGGREGATION_METHOD, METRIC_MISSING} = STRINGS;
+const {INVALID_AGGREGATION_METHOD} = STRINGS;
 
 /**
  * Validates metrics array before applying aggregator.
  * If aggregation method is `none`, then throws error.
  */
-const checkIfMetricsAreValid = (metrics: string[], parameters: Parameters) => {
-  metrics.forEach(metric => {
-    const method = getAggregationMethod(metric, parameters);
-
+const checkIfMetricsAreValid = (context: ManifestCommon) => {
+  context.aggregation?.metrics.forEach(metric => {
+    const method = getAggregationMethod(metric, PARAMETERS);
     if (method === 'none') {
       throw new InvalidAggregationParams(INVALID_AGGREGATION_METHOD(method));
     }
   });
 };
 
-/**
- * Aggregates child node level metrics. Validates if metric aggregation type is `none`, then rejects with error.
- * Otherwise iterates over inputs by aggregating per given `metrics`.
- */
-export const aggregate = (
-  inputs: PluginParams[],
-  metrics: string[],
-  parameters: Parameters
-) => {
-  checkIfMetricsAreValid(metrics, parameters);
+/**Returns Boolean indicating whether a node has children */
+function hasChildren(node: Node) {
+  if (node['children']) {
+    return true;
+  }
+  return false;
+}
 
-  return inputs.reduce((acc, input, index) => {
-    for (const metric of metrics) {
-      if (!(metric in input)) {
-        throw new InvalidAggregationParams(METRIC_MISSING(metric, index));
+/** Returns children of given node */
+function getChildren(node: Node): Node[] {
+  const children: Node[] = [];
+  Object.keys(node.children).forEach(child => {
+    children.push(node.children[child]);
+  });
+  return children;
+}
+
+/** element-wise addition of two arrays */
+function aggregateArrays(a: Array<any>, b: Array<any>, method: string) {
+  if (method === 'sum') {
+    return a.map((element, idx) => {
+      return element + b[idx];
+    });
+  } else if (method === 'avg') {
+    return a.map((element, idx) => {
+      return (element + b[idx]) / 2;
+    });
+  }
+  return a;
+}
+
+/** aggregates multiple time series into one single time series*/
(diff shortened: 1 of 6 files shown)


src/lib/aggregator.ts
```diff
@@ -1,58 +1,144 @@
...
+  } else if (method === 'avg') {
+    return a.map((element, idx) => {
+      return (element + b[idx]) / 2;
+    });
+  }
+  return a;
+}
+
+/** aggregates multiple time series into one single time series*/
+export const aggregate = (tree: Node, context: ManifestCommon) => {
```
manushak, line 56: please consider breaking down the logic of the function into smaller functions? It would be great if you could create functions named `aggregateVertical` and `aggregateHorizontal`, and distribute the corresponding logic accordingly within them.

src/lib/aggregator.ts
```diff
@@ -1,58 +1,144 @@
...
+  // check if root has children
+  if (hasChildren(tree)) {
+    // iterate through child nodes
+    getChildren(tree).forEach(node => {
+      // if the child has children, start recursive call to aggregate()
+      if (hasChildren(node)) {
+        aggregate(node, context);
       }
+      context.aggregation?.metrics.forEach(metric => {
+        let treeTimeSeries = tree.inputs?.map(input => input[`${metric}`]);
```
jawache, line 67: So tree here is referring to the parent node of the current node.

src/lib/aggregator.ts
```diff
@@ -1,58 +1,144 @@
...
 
-    return acc;
-  }, {} as AggregationResult);
+        if (
+          context.aggregation?.type === 'both' ||
+          context.aggregation?.type === 'vertical'
+        ) {
+          if (tree[`aggregated-${metric}`]) {
+            if (node[`aggregated-${metric}`]) {
+              tree[`aggregated-${metric}`] = aggregateArrays(
```
jawache, line 91: I _think_ what we are doing here would only work if a parent node had 1 child since the aggregateArrays only aggregates two arrays together.

More specifically i think this logic only works if the number of children is > 1 AND the aggregation method is sum since it will just keep on adding the values to the parent node.
OR
If the aggregation method is average it would only work if there is 1 child since the seconds childs average would somehow overwrite the average figure from the first child, i don't think this would average all the children values and put them on the parent.

We also need to support other aggregation options, I can imagine scenarios where we just pick the first one in the list, or maybe do some string manipulation if you are aggregating a descriptive field. Unsure, but it only really works I _think_ if you are aggregating all the children in one go with each other so you know all the values.

src/lib/aggregator.ts
```diff
@@ -1,58 +1,144 @@
...
+                tree[`aggregated-${metric}`],
```
jawache, line 92: I'm unclear when `tree[`aggregated-${metric}`]` would be set?

src/lib/aggregator.ts
```diff
@@ -1,58 +1,144 @@
...
+function getChildren(node: Node): Node[] {
+  const children: Node[] = [];
+  Object.keys(node.children).forEach(child => {
+    children.push(node.children[child]);
+  });
+  return children;
+}
+
+/** element-wise addition of two arrays */
+function aggregateArrays(a: Array<any>, b: Array<any>, method: string) {
```
jawache, line 42: I'm unclear how this would work if you had a node with 4 children and you had to sum (or avg, or pick-first, or another selection criteria)?

src/lib/aggregator.ts
```diff
@@ -1,58 +1,144 @@
...
+        let treeTimeSeries = tree.inputs?.map(input => input[`${metric}`]);
 
-      const accessKey = `aggregated-${metric}`;
-      acc[accessKey] = acc[accessKey] ?? 0;
-      acc[accessKey] += parseFloat(input[metric]);
+        if (
+          treeTimeSeries === undefined ||
+          treeTimeSeries?.includes(undefined)
+        ) {
+          treeTimeSeries = new Array(3).fill(0);
```
jawache, line 73: Ahh I see, so if the parent node doesn't have inputs we just create zeros (but specifically an array of 3?) and this would make the aggregate function work logically.

jawache (changes requested):
Hey @jmcook1186 I can see how the assumption that nodes can be both component and grouping nodes sent you down this path and the complexity of the solution! The parent pointers make a lot more sense to me now.

I suspect given my review of the code that it currently only works for sum based aggregations and would break (not be accurate) for avg based.

But given the complexity I do think that if we can insist that grouping nodes have no inputs, then the solution becomes a lot simpler and easier to understand (and extend).

Can we come back to this and implement using the alternative algorithm I proposed with an output format aligned to what I proposed for the CSV exporter (https://github.com/Green-Software-Foundation/if/issues/411). It would make everything snap into place with simplicity?

src/lib/aggregator.ts
jmcook1186: yes, exactly the 3 is a typo - it should be grabbbing the length of the node input array

src/lib/aggregator.ts
jmcook1186: yes, you can swap "tree" for "root" if it's easier to reason about

jmcook1186: agree that the ;logic will be much easier if the parent cannot also have its own input data

src/lib/aggregator.ts
```diff
@@ -1,58 +1,144 @@
...
+        }
+
+        let nodeTimeSeries = node.inputs?.map(input => input[`${metric}`]);
```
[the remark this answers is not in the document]
jmcook1186: yes

jmcook1186:
closing as superseded by @narekhovhannisyan 's version

PR #441, opened 2024-02-18
exhaust: added exhaust function, with stub for exhaust-csv-export and…

pazbardanl:
… a sample impl

### Types of changes

- added exhaust function
- added csv-exporter exhaust plugin


src/models/exhaust-csv-export.ts
```diff
@@ -0,0 +1,54 @@
+import * as path from 'path';
+import * as fs from 'fs/promises';
+import {ERRORS} from '../util/errors';
+import {ExhaustPluginInterface} from './exhaust-plugin';
+const {InputValidationError} = ERRORS;
+
+export class ExhaustCsvExporter implements ExhaustPluginInterface {
```
narekhovhannisyan, line 7: we should use functional approach, let's go through tomorrow

pazbardanl: I know, it's a stub (that's why the PR is default)

src/lib/exhaust.ts
```diff
@@ -0,0 +1,31 @@
+import {ExhaustExportCsv} from '../models/exhaust-export-csv';
+
+const createExhaustPlugin = (pluginTypeName: string) => {
+  switch (pluginTypeName) {
```
narekhovhannisyan, line 4: we should discuss if plugins are another repository or builtins

src/models/exhaust-plugin-interface.ts
```diff
@@ -0,0 +1,6 @@
+export interface ExhaustPluginInterface {
```
narekhovhannisyan, line 1: this can me moved to types folder

src/models/index.ts
```diff
@@ -1,2 +1,3 @@
 export {GroupBy} from './group-by';
 export {TimeSync} from './time-sync';
+export {ExhaustExportCsv} from './exhaust-export-csv';
```
narekhovhannisyan, line 3: can be renamed to ExportCsv

src/models/exhaust-export-csv.ts
```diff
@@ -0,0 +1,121 @@
...
+    const outputPath: string =
+      'output-path' in config
+        ? config['output-path']
+        : (() => {
+            throw new InputValidationError("Config does not have 'outputPath'");
+          })();
+    return [outputPath];
+  };
+
+  const handleLeafValue = (
```
narekhovhannisyan, line 17: please provide JSDoc style documentation for each method to make clear what function does without reading the code.

src/lib/exhaust.ts
```diff
@@ -0,0 +1,31 @@
...
+    case 'csv':
+      return ExhaustExportCsv(pluginConfigItems);
+    default:
+      throw new Error(`unkonwn exhaust plugin type: ${pluginTypeName}`);
+  }
+};
+
+export const exhaust = (context: any, tree: any) => {
+  if (context && context.initialize.outputs) {
```
narekhovhannisyan, line 23: you can check only for outputs, and receive outputs instead of context.
Since all other methods are parsing tree as a first argument, then options exhaust can follow the same rules

src/lib/exhaust.ts
narekhovhannisyan: I think we got approval to import exhaust plugins as we are doing with ordinary plugins

src/models/index.ts
pazbardanl: Fine by me, just wanted to make sure we don't to have some convention to differentiate regular plugins from exhaust plugins

src/lib/exhaust.ts
pazbardanl: in retrospect i think this makes sense.
looking at the way ordinary plugins are being created from the manifest, i think applying a similar flow for creation of the exhaust plugins means significant changes. i still think we should do it :) but maybe not in the scope of this task.
i propose maybe pushing the csv-export exhaust functionality as it is in this PR, and create a new issue (with the relevant priority) for refactoring it into external plugins. WDYT?

narekhovhannisyan (approved):

PR #487, opened 2024-03-02
feat: turn impact-engine into a more comprehensive cli tool

alexzurbonsen:
### Types of changes

This PR is a suggestion to create a more comprehensive CLI tool that is able to add and remove plugins to itself thus avoiding global installs.

Background: We have been working on issues of the Carbon Hack and didn't find the approach of installing all plugins globally too fortunate. Thus, here is a suggestion for an alternative approach.

This PR puts a `plugins` directory next to `if` where all plugins can be installed.
To that end it adds three new commands
```
ie add <plugin>
ie remove <plugin>
ie list
```
As before there is still 
```
ie --manifest <path>
# or
ie run --manifest <path>
```

There are a couple of things that would still have to be considered if this PR would actually find acceptance
- documentation changes
- further tests, especially of the main method
- a couple of docstrings are still missing I think


alexzurbonsen:
This PR basically moved every file. So it will constantly have merge conflicts. I will of course solve them but before I invest more time in this I wanted to get some feedback from you guys.

if/src/lib/initialize.ts
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  return undefined;
+};
+
+/**
+ * Imports module by given `path`.
+ */
+const importModuleFrom = async (path: string) => {
+  const isModuleInstalled = await checkIfModuleIsInstalled(path);
```
alexzurbonsen, line 70: This doesn't seem like a good idea anymore. It just adds complexity and the current implementation opens up the tool to shell injection attacks (even though from a manifest file that you decide to execute yourself - but still unnecessary I think).

It should be enough to check if `resolve(pluginsDir, path)` is still in `plugins`.

jmcook1186:
Hi @alexzurbonsen thanks for the PRs - we'll chat about this idea in our team call this morning and get back to you shortly. (cc @narekhovhannisyan)

jmcook1186:
Hi @alexzurbonsen we triaged this today and we agree there are some interesting things here that we might like to merge in to IF. However, there's a lot here and we think we'll want to cherry pick a bit and work with you on bringing it in. 

If it works for you, we'll build in some time to spike on this internally so we can get you some more specific feedback and go from there. 

We want to be extra careful with major changes at the moment because we're just one week away from our pre-hackathon code freeze. cc @narekhovhannisyan

alexzurbonsen:
@jmcook1186 Thank you for the quick feedback. Sure, sounds like a good approach.

I totally understand you want to be careful before the hackathon...

I will hopefully go over it in the evening once more to take out this module installation check again.

alexzurbonsen:
I changed the approach a bit. Not, `ie` doesn't output npm's stdout for the relevant commands but produces its own output (and only passes on npm output in case of errors).

It looks like that (almost, the screenshot wasn't made with the latest version. Some punctuation changes were made.)

ie

jmcook1186:
Hi @alexzurbonsen just to note we haven't forgotten about this PR. It's a lot of changes and we need to schedule some time to really go through and work out which bits we want, and then probably create some individual tickets for it. I anticipate this happening in a couple of weeks when we start working on a "developer experience" epic, until then we'll focus on reviewing your other PRs across the IF and plugin repos.

Cheers

alexzurbonsen:
Hi @jmcook1186, sounds good. Thanks four your message.

jmcook1186:
Hi @alexzurbonsen thanks for this, there are some ideas in here we will probably want to return to later but right now we are building in a slightly different direction and better to close this out and revisit in the future. In particular, we like the idea of `ie add` for managing dependencies but we're not likely to bring in a big change to the framework right now apart from what is on our roadmap.

PR #516, opened 2024-03-09
optimize merge defaults strategy

alexzurbonsen:
### Types of changes

I wrote a test suite that explains my understanding of what the function should do. The current `mergeObject` function fails this suite. E.g. it overrides an input if its value is false and the default is true. It does not deep merge objects.

Changed the implementation to pass the tests.


src/lib/compute.ts
```diff
@@ -21,7 +21,7 @@ const mergeDefaults = (
 ) => {
   if (inputs) {
     const response = defaults
-      ? inputs.map(input => mergeObjects(input, defaults))
+      ? inputs.map(input => mergeObjects({...defaults}, input))
```
MariamKhalatova, line 24: Hi @alexzurbonsen , thanks for your contribution. 
Just wandering if there is specific reason for using object destructuring on defaults?

src/util/helpers.ts
```diff
@@ -2,6 +2,7 @@ import {ERRORS} from './errors';
 
 import {STRINGS} from '../config';
 import {logger} from './logger';
+import {mergeWith} from 'lodash';
```
narekhovhannisyan, line 5: btw can you please move 3rd party library import to the top?

```
import {mergeWith} from 'lodash';

import {ERRORS} from './errors';
import {logger} from './logger';

import {STRINGS} from '../config';
```

src/__tests__/unit/util/helpers.test.ts
```diff
@@ -0,0 +1,119 @@
...
+      };
+
+      const result = mergeObjects(defaults, input);
+      const expectedResult = {
+        a: 1,
+        b: false,
+        c: 'testInput',
+      };
+
+      expect(result).toEqual(expectedResult);
```
MariamKhalatova, line 25:
```ts
const expectedResult = {
        a: 1,
        b: false,
        c: 'testInput',
      };
```
This expected result can be removed and input can be directly used in assertion, f.ex. `expect(result).toEqual(input);`

src/lib/compute.ts
alexzurbonsen: Hi @MariamKhalatova, that is because the function uses lodash's `mergeWith` function. And that writes `input` into `defaults`. That is why we need to copy the top level of `defaults`, I think.

alexzurbonsen: The docs: [mergeWith](https://lodash.com/docs/4.17.15#mergeWith) and [merge](https://lodash.com/docs/4.17.15#merge)

alexzurbonsen: I added a test, that might help understand the issue. If you replace `{...defaults}` with `defaults` and run the test, you can see how the object reference kicks in.

I add to export `mergeDefaults` to make this testable. But I guess its worth it.

alexzurbonsen:
Thank you for your comments. I adressed them.

narekhovhannisyan:
Hi @alexzurbonsen, we have gone through this PR on our weekly triage. Everything seems good and we are keen to merge your PR if it can be rewritten without `lodash` library since the code uses only one function from it.

alexzurbonsen:
Hi @narekhovhannisyan, to be honest I am not a fan of this approach in this case. Writing this kind of code is fairly complex or at least complex enough that you easily make mistakes or overlook edge cases. In this kind of situation I would prefer to use a library. The function we are fixing in this PR is a good example, why it isn't a good idea to write this stuff yourself.

Thus, I won't invest time to rewrite a lodash function that works better than any code I will be writing myself.

alexzurbonsen:
But of course, you can take over the PR if you like.

narekhovhannisyan:
@alexzurbonsen we decided to not merge arrays and nested objects to keep everything straightforward. Will push changes for tests and the implementation accordingly.

narekhovhannisyan (approved):

src/util/helpers.ts
```diff
@@ -20,15 +18,20 @@ export const andHandle = (error: Error) => {
...
-    return;
-  };
 
-  return mergeWith(destination, source, handleArrays);
+    if (!(key in input)) {
+      merged[key] = defaults[key];
+    }
+  }
+
+  return merged;
```
alexzurbonsen, line 36: Why not just use `{...defaults, ...inputs}` now?

alexzurbonsen:
> @alexzurbonsen we decided to not merge arrays and nested objects to keep everything straightforward. Will push changes for tests and the implementation accordingly.

I guess it might be helpful if this behaviour is documented somewhere?

src/util/helpers.ts
narekhovhannisyan: it won't do the trick since null values won't be overridden

narekhovhannisyan: and that behavior should be covered by tests too, the current implementation doesn't satisfy that either

MariamKhalatova (approved):

PR #529, opened 2024-03-14
Update Readme file

manushak:
### Types of changes

- update readme file
- remove credentials of watt-time plugin in basic-demo manifest
- update parameters in basic-demo manifest

### DIFF
--- a/src/builtins/README.md
+++ b/src/builtins/README.md
@@ -5,6 +5,7 @@ Time sync standardizes the start time, end time and temporal resolution of all o
 ## Parameters
 
 ### Plugin config
+
 The following should be defined in the plugin initialization:
 
 - `start-time`: global start time as ISO 8061 string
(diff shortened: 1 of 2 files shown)


jmcook1186 (approved):
ok, lgtm!

src/builtins/README.md
```diff
@@ -58,7 +58,7 @@ This looks as follows:
 ]
 ```
 
-Each `observation` actually includes many key-value pairs. The precise content of the `observation` is not known until runtime because it depends on which plugins have been included in the pipeline. Different values have to be treated differently when we upsample in time. The method we use to upsample depends on the `aggregation-method` defined for each key in `units.yml`. 
+Each `observation` actually includes many key-value pairs. The precise content of the `observation` is not known until runtime because it depends on which plugins have been included in the pipeline. Different values have to be treated differently when we upsample in time. The method we use to upsample depends on the `aggregation-method` defined for each key in `units.yml`.
```
narekhovhannisyan, line 61: should it be observation or inputs?

jmcook1186: it could be `input`. I';ve treated `observation` as the object, rather than using `observation` to refer to individual key/value pairs. wyt?

narekhovhannisyan: yeah could be. Observation explains better the meaning

narekhovhannisyan (approved):

PR #578, opened 2024-04-05
Add unit test for args.ts

manushak:
### Types of changes

- add full test coverage for args.ts
-

### DIFF
--- a/src/__tests__/unit/util/args.test.ts
+++ b/src/__tests__/unit/util/args.test.ts
@@ -5,7 +5,7 @@ jest.mock('ts-command-line-args', () => ({
   parse: () => {
     switch (process.env.result) {
       case 'throw-error-object':
-        throw new CliInputError(MANIFEST_IS_MISSING);
+        throw new Error(MANIFEST_IS_MISSING);
       case 'error':
         throw MANIFEST_IS_MISSING;
       case 'manifest-is-missing':
--- a/src/__tests__/unit/util/args.test.ts
+++ b/src/__tests__/unit/util/args.test.ts
@@ -1,20 +1,36 @@
+const processRunningPath = process.cwd();
+
 jest.mock('ts-command-line-args', () => ({
   __esModule: true,
   parse: () => {
     switch (process.env.result) {
+      case 'throw-error-object':
+        throw new CliInputError(MANIFEST_IS_MISSING);
       case 'error':
+        throw MANIFEST_IS_MISSING;
+      case 'manifest-is-missing':
         return {};
       case 'manifest':
         return {
           manifest: 'manifest-mock.yml',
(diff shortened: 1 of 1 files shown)


src/__tests__/unit/util/args.test.ts
```diff
@@ -35,28 +51,47 @@ import path = require('path');
...
   const originalEnv = process.env;
 
   describe('parseArgs(): ', () => {
     it('throws error if there is no argument passed.', () => {
-      expect.assertions(2);
+      expect.assertions(5);
 
       process.env.result = 'error'; // used for mocking
 
       try {
```
narekhovhannisyan, line 73: shouldn't try-catch blocks be in separate unit tests?

manushak: I've combined all of them because the meaning is the same

src/__tests__/unit/util/args.test.ts
```diff
@@ -1,20 +1,36 @@
+const processRunningPath = process.cwd();
+
 jest.mock('ts-command-line-args', () => ({
   __esModule: true,
   parse: () => {
     switch (process.env.result) {
+      case 'throw-error-object':
+        throw new CliInputError(MANIFEST_IS_MISSING);
       case 'error':
+        throw MANIFEST_IS_MISSING;
```
narekhovhannisyan, line 10: do we need to support these two cases?
both are throwing error, one with error class, the other without

src/__tests__/unit/util/args.test.ts
```diff
@@ -1,20 +1,36 @@
+const processRunningPath = process.cwd();
+
 jest.mock('ts-command-line-args', () => ({
   __esModule: true,
   parse: () => {
     switch (process.env.result) {
+      case 'throw-error-object':
+        throw new CliInputError(MANIFEST_IS_MISSING);
```
narekhovhannisyan, line 8: btw it's not the parse function's responsibility to throw CliInputError

src/__tests__/unit/util/args.test.ts
manushak: yeah, because of `validateAndParseProcessArgs` function:
```
const validateAndParseProcessArgs = () => {
  try {
    return parse<ManifestProcessArgs>(ARGS);
  } catch (error) {
    if (error instanceof Error) {
      throw new CliInputError(error.message);
    }

    throw error;
  }
};
```

narekhovhannisyan: ok you're covering here `throw error;` case

narekhovhannisyan (approved):

MariamKhalatova (approved):

Issue #613, opened 2024-04-10
Revise issue refinement/readiness process


jmcook1186:
Sub of #651 

## User story
As a core developer I want to treat issues consistently and make sure I know what changes are required to close a ticket in advance of working on it. I also want to be able to spec tickets sufficiently so that community contributors can work on them with clarity and with unambiguous expectations.

## Rationale
We currently work on issues based on internal conversations or loose specifications, and sometimes we are guilty of defining the task while it is in flight. This doesn't serve the community very well as it is a barrier to contributing effectively, and it has also led to miscommunications within the team. Therefore, we should inspect our processes and devise something better.

## Implementation details
- [x] examine current process
- [x] propose changes, probably including a revised issue template
- [ ] document issue readiness criteria
- [ ] establish any additional team calls, calendar events or other scheduling we need to ensure tickets get refined

## Priority
4/5

## Size
M

## What does "done" look like?
Process updated, documented and implemented

## Deadline
tbc

jmcook1186:
Hi @narekhovhannisyan @manushak @MariamKhalatova @jawache @pazbardanl here is a suggestion for a ticket refinement/readiness process to implement across IF repositories. Please use this as a starting point for discussions and provide feedback in comments here!

## Refinement

Refinement is the process of taking a loosely scoped or ambiguous ticket and making it so well-described that anyone could pick it up off the board and produce an equivalent outcome to one of the core team. It's important we do this diligently even if it feels unnecessary or frictionful, because:

- it keeps us all aligned with the purpose and rationale of individual tasks
- it improves our transparency
- it lowers the barrier to entry for contributors
- it helps us to "measure twice, cut once" and avoid doing work more than once
- it helps us to think together about each issue
- it provides an archive of our thinking that we can go back to in future.

We can make the refinement process as consistent as possible by sticking faithfully to issue templates. By choosing the fields carefully and filling them out thoroughly, we can ensure the right information is available, and using checkboxes for subtask will help us to track progress and be precise about the scope of the ticket.

## Peer-review

We will also implement a mini peer-review process for each ticket, where a team member is required to review a ticket and make it READY - we should not typically make our own tickets READY. Instead, when you create a ticket, you assign another team member to be the reviewer. Then, responsibility for the thoroughness of the information is shared and we can hold each other accountable.

The reviewer can add the `READY` label when they are satisfied with the ticket quality.

## Suggested Issue Content

- **rationale/context:** why is this change required, what are the intended outcomes
[...]

PR #642, opened 2024-04-12
update dev ticket template

jmcook1186:

Updates developer ticket template according to discussion here: https://github.com/Green-Software-Foundation/if/issues/613

### DIFF
--- a/.github/ISSUE_TEMPLATE/dev-ticket.md
+++ b/.github/ISSUE_TEMPLATE/dev-ticket.md
@@ -8,7 +8,7 @@ assignees: ''
 ---
 
 ## User story
-<!--- Describe the user persona this issue affects and how they benefiot from the proposed changes -->
+<!--- Describe the user persona this issue affects and how they benefit from the proposed changes -->
 
 ## Rationale/context
 <!--- Explain any relevant background and describe why the proposed changes are needed -->
--- a/.github/ISSUE_TEMPLATE/dev-ticket.md
+++ b/.github/ISSUE_TEMPLATE/dev-ticket.md
@@ -35,11 +35,10 @@ assignees: ''
 <!--- list the specific tasks that need to be completed in order to achieve the aims of this ticket. Please use checkboxes!
 
 Please use the given-when-then format to describe the acceptance criteria:
-
-Given (Setup): Describes the initial state of the system or the preconditions for a change.
-
-When (Action): Describes the specific action or behavior that is being tested/changed.
-
-Then (Assertion): Defines the expected outcome or behavior of the system after the action in the "When" step is performed.
-
 -->
+
+**Given** (Setup): Describes the initial state of the system or the preconditions for a change.
+
+**When** (Action): Describes the specific action or behavior that is being tested/changed.
+
+**Then** (Assertion): Defines the expected outcome or behavior of the system after the action in the "When" step is performed.
--- a/.github/ISSUE_TEMPLATE/dev-ticket.md
+++ b/.github/ISSUE_TEMPLATE/dev-ticket.md
@@ -7,35 +7,38 @@ assignees: ''
 
 ---
 
-<!--- Provide a general summary of the issue in the Title above -->
+## User story
+<!--- Describe the user persona this issue affects and how they benefiot from the proposed changes -->
 
-## Rationale
-<!--- Provide the relevant background information to udnerstand why this change is needed -->
-<!--- This is where you can provide more details thathelp the reader understand why you want to make this change -->
+## Rationale/context
+<!--- Explain any relevant background and describe why the proposed changes are needed -->
 
-## Implementation details
-<!--- Explain the steps you will take to implement this change. If you do not have specific steps yet, write your implementation ideas or outline general approach -->
+## Implementation details:
+<!--- what specific steps will be taken to address the aims of the ticket? -->
 
-## Priority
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/dev-ticket.md
```diff
@@ -7,35 +7,38 @@ assignees: ''
 
 ---
 
-<!--- Provide a general summary of the issue in the Title above -->
+## User story
+<!--- Describe the user persona this issue affects and how they benefiot from the proposed changes -->
```
manushak, line 11: typo in `benefiot`

.github/ISSUE_TEMPLATE/dev-ticket.md
```diff
@@ -7,35 +7,38 @@ assignees: ''
...
+<!--- how long will this take (t-shirt sizing XS, S, M, L, XL) -->
+
+## Prerequisites/resources
+<!--- does anything else need to happen to enable this work to be completed? -->
+
+## Updates to documentation
+<!--- enumerate specific changes required to the documentation, formatted as checkboxes with associated urls -->
+
+## Testing
+<!--- what tests are required for the changes to get merged? Please use checkboxes!-->
```
narekhovhannisyan, line 32: I think it's better to leave at least one line of checkboxes so you don't have to figure out how to write it

narekhovhannisyan (approved):

.github/ISSUE_TEMPLATE/dev-ticket.md
jmcook1186:
```suggestion
<!--- Describe the user persona this issue affects and how they benefit from the proposed changes -->
```

.github/ISSUE_TEMPLATE/dev-ticket.md
jmcook1186: good point - adding now

jawache:
Thanks @jmcook1186, every project is now being asked to align to the pm protocol here: https://www.notion.so/grnsft/Internal-Project-Protocol-6b8a9e881ed942df9a81a60aeecfd998?pvs=4 head to the section called "issues". So this will need to change somewhat to conform.

Specifically for IF the ticket should look something like so which follows the protocol and merges with your suggestions.

**What**: one sentence to describe the deliverable.
**Why**: can use a user story, but let's not insist, if it's a smaller subtask then user stories just become repetitive and add limited value (as a developer the tests should pass so we have a high quality product) so I think there is room for flexibility.
**Context**: As per your suggestion 
**Prerequisites**: As per your suggestion 
**SoW**: replaces implementation detail, docs and testing, can add comments to encourage people to think of checklist items related to testing/docs. But this is the place with all the work that needs to be done.
**Acceptance Criteria**: this should replace "What does done look like?" if the ticket involves a manifest file one of the criteria should be a manifest file that should be runnable. If there is any testing to be done (automation or manual) then the acceptance criteria is the place to understand what testing needs to be done. I'd imagine we flesh this out in refinement with multiple heads.

In addition 


- Size should be either a project board field or a label, we might insist on one or the other in the future but right now there is flexibility to choose. In fact maybe not something we insist on until we've done some refinement on a ticket? But regardless, label or board field, you can only fill it in after the ticket is created.

- We normally keep it simple with just bolded words for headings (see hackathon board for examples) but I think we can also use headings instead of bolded but let's stick to h4, h2s take up a lot of space unnecessarily 😁

jawache (changes requested):
As per my comment on the PR

narekhovhannisyan (approved):

manushak (approved):

.github/ISSUE_TEMPLATE/dev-ticket.md
```diff
@@ -7,35 +7,29 @@ assignees: ''
...
-
-## Does this require updates to documentation or other materials??
-<!--- Are there documentation, website or other media changes required after this change is implemented?. -->
-
-## Deadline
-<!--- Does this need to happen by a specific date? -->
-
-## Acceptance criteria
-<!--- What has to be true for this to get merged? What threshold of testing is acceptable? -->
+**Acceptance criteria**
```
jawache, line 31: @jmcook1186 let's add to the template to trigger thinking in that style 

**Given** ...
**When** ...
**Then** ...

.github/ISSUE_TEMPLATE/dev-ticket.md
```diff
@@ -7,35 +7,29 @@ assignees: ''
...
-## Priority
-<!--- How urgent is this issue, and why? -->
-<!--- Explain in writing or give a numerical value from 1 (nice-to-have) to 5 (critical) -->
+**Prerequisites/resources**
+<!--- does anything else need to happen to enable this work to be completed? -->
 
-## Scope
-<!--- Which parts of the project/codebase does this affect? -->
-<!--- What are the possibel downstream impacts or side-effects of this change?? -->
+**Statement of Work**
```
jawache, line 22: I believe the title is SoW (Scope of Work?)

It should also be a checklist, this is the set of tasks that need to be completed and if the task is complex it might need to turn into a linked issue.

Also recommend adding in some predefined checklist items such as 

- [ ] if-docs updated 
- [ ] test cases added

jawache (changes requested):

jawache (approved):

PR #732, opened 2024-05-21
Create a generic Interpolation plugin

manushak:
### Types of changes

- Create a generic interpolation plugin
-


jmcook1186 (changes requested):
Hi @manushak 
thanks for working on this but it doesn't meet the issue spec. The plugin should be generic, by which I mean it should interpolate any arbitrary data that is passed to it in the way described in the acceptance criteria in the ticket, and not require specific parameters such as `vcpus-allocated`. Please check the ticket description to see how the plugin should work.

manushak:
> Hi @manushak thanks for working on this but it doesn't meet the issue spec. The plugin should be generic, by which I mean it should interpolate any arbitrary data that is passed to it in the way described in the acceptance criteria in the ticket, and not require specific parameters such as `vcpus-allocated`. Please check the ticket description to see how the plugin should work.

@jmcook1186  I've updated, please review

jmcook1186 (approved):
Pushed some fixes - now satisfied this works as expected. Added some new validations and updated some error messages for clarity, and added some tests.

jmcook1186:
over to @narekhovhannisyan for second review

examples/manifests/interpolation.yml
```diff
@@ -0,0 +1,26 @@
...
+tree:
+  children:
+    child:
+      pipeline:
+        - interpolation
+      inputs:
+        - timestamp: 2023-07-06T00:00
+          duration: 3600
+          cpu/utilization: 45
+          # vcpus-allocated: 4
```
narekhovhannisyan, line 25: should they stay commented?

src/util/validations.ts
```diff
@@ -83,21 +83,22 @@ export const manifestSchema = z.object({
  * Validates given `manifest` object to match pattern.
  */
 export const validateManifest = (manifest: any) =>
-  validate(manifestSchema, manifest, ManifestValidationError);
+  validate(manifestSchema, manifest, undefined, ManifestValidationError);
```
narekhovhannisyan, line 86: why undefined?

examples/manifests/interpolation.yml
manushak: thanks for noticing, the manifest moved to the manifests/plugins folder

src/util/validations.ts
manushak: The `validate` function is used by the `validateManifest` function, which is called from plugins too. To avoid creating a new function, I assigned `index` to `undefined` here.

narekhovhannisyan: in that case either make it last argument, or pass all the info to validate function as a single object

manushak: Both options are not achievable because a plugin calls this version of the validate function: validate<z.infer<typeof schema>>(schema, input, index). If we change it to a single object, all plugins will need to be updated accordingly. If we change the order of the arguments, we would need to provide an Error class in every plugin. I hope I have described the situation :)

PR #753, opened 2024-05-31
Improve IF logging protocols

manushak:
### Types of changes

- add `--debug` command functionality
- add debug log messages into the codebase accordingly
-

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -15,7 +15,7 @@ import {validateManifest} from './util/validations';
 
 import {STRINGS} from './config';
 
-const {DISCLAIMER_MESSAGE, EXITING_IF, STARTING_IMPACT_FRAMEWORK} = STRINGS;
+const {DISCLAIMER_MESSAGE, EXITING_IF, STARTING_IF} = STRINGS;
 
 const impactEngine = async () => {
   const options = parseIEProcessArgs();
@@ -24,7 +24,7 @@ const impactEngine = async () => {
   debugLogger.overrideConsoleMethods(!!debug);
 
   logger.info(DISCLAIMER_MESSAGE);
-  console.info(STARTING_IMPACT_FRAMEWORK);
+  console.info(STARTING_IF);
 
   const {rawManifest, parameters} = await load(inputPath, paramPath);
   const envManifest = await injectEnvironment(rawManifest);
--- a/src/index.ts
+++ b/src/index.ts
@@ -7,6 +7,7 @@ import {initialize} from './lib/initialize';
 import {load} from './lib/load';
 import {parameterize} from './lib/parameterize';
 
+import {debugLogger} from './util/debug-logger';
 import {parseIEProcessArgs} from './util/args';
 import {andHandle} from './util/helpers';
 import {logger} from './util/logger';
@@ -14,13 +15,16 @@ import {validateManifest} from './util/validations';
 
 import {STRINGS} from './config';
 
-const {DISCLAIMER_MESSAGE} = STRINGS;
+const {DISCLAIMER_MESSAGE, STARTING_IMPACT_FRAMEWORK, EXITING_IF} = STRINGS;
 
 const impactEngine = async () => {
   const options = parseIEProcessArgs();
+  const {inputPath, paramPath, outputOptions, debug} = options;
+
+  debugLogger.overrideConsoleMethods(!!debug);
 
   logger.info(DISCLAIMER_MESSAGE);
-  const {inputPath, paramPath, outputOptions} = options;
+  console.info(STARTING_IMPACT_FRAMEWORK);
 
   const {rawManifest, parameters} = await load(inputPath, paramPath);
   const envManifest = await injectEnvironment(rawManifest);
@@ -44,6 +48,7 @@ const impactEngine = async () => {
       }
     }
   }
+  console.info(EXITING_IF);
 };
 
 impactEngine().catch(andHandle);
--- a/src/index.ts
+++ b/src/index.ts
@@ -15,7 +15,7 @@ import {validateManifest} from './util/validations';
 
 import {STRINGS} from './config';
 
-const {DISCLAIMER_MESSAGE, STARTING_IMPACT_FRAMEWORK, EXITING_IF} = STRINGS;
+const {DISCLAIMER_MESSAGE, EXITING_IF, STARTING_IMPACT_FRAMEWORK} = STRINGS;
 
 const impactEngine = async () => {
   const options = parseIEProcessArgs();
--- a/src/lib/parameterize.ts
+++ b/src/lib/parameterize.ts
@@ -1,23 +1,34 @@
 import {logger} from '../util/logger';
 import {memoizedLog} from '../util/log-memoize';
+import {debugLogger} from '../util/debug-logger';
 
 import {STRINGS, PARAMETERS} from '../config';
 
 import {ManifestParameter} from '../types/manifest';
 import {Parameters} from '../types/parameters';
 
-const {REJECTING_OVERRIDE, UNKNOWN_PARAM} = STRINGS;
+const {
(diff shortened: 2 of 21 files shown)


jmcook1186 (changes requested):
Seems like something is not right here, see these logs for example:
```
DEBUG: 2024-05-31T12:29:04.237Z: teads-curve: Computing pipeline for `sum`
DEBUG: 2024-05-31T12:29:04.238Z: sum: Computing pipeline for `coefficient`
DEBUG: 2024-05-31T12:29:04.238Z: coefficient: Computing pipeline for `multiply`
```

the plugin names do not match (e.g. **teads-curve**: Computing pipeline for **sum**)

Also I don't see the `console.x` rerouting working. Can you confirm?

jmcook1186 (approved):
lgtm - tested with two manifests from our examples folder and saw the logs I exected when running using `--debug`. Also tried injecting some dummy log statements into the builtins and saw their messages get emitted via the IF logger during the execution, which is great.

Thanks

src/lib/parameterize.ts
```diff
@@ -1,23 +1,34 @@
...
 
 import {ManifestParameter} from '../types/manifest';
 import {Parameters} from '../types/parameters';
 
-const {REJECTING_OVERRIDE, UNKNOWN_PARAM} = STRINGS;
+const {
+  REJECTING_OVERRIDE,
+  UNKNOWN_PARAM,
+  SYNCING_PARAMETERS,
+  // CHECKING_AGGREGATION_METHOD,
```
narekhovhannisyan, line 14: ?

src/lib/parameterize.ts
```diff
@@ -1,23 +1,34 @@
...
+  console.debug(SYNCING_PARAMETERS);
+
   let parametersStorage = PARAMETERS;
 
   /**
    * Returns aggregation method for given `unitName`. If doesn't exist then returns value `sum`.
    */
   const getAggregationMethod = (unitName: string) => {
+    debugLogger.removeExecutedPluginName();
+    // console.debug(CHECKING_AGGREGATION_METHOD(unitName));
```
narekhovhannisyan, line 30: ?

src/lib/parameterize.ts
manushak: thanks for noticing. I forgot uncomment before push

narekhovhannisyan:
Screenshot 2024-06-03 at 17 48 01

this log appear whenever if is executed without adding --debug flag.

src/index.ts
```diff
@@ -7,20 +7,24 @@ import {initialize} from './lib/initialize';
...
+import {debugLogger} from './util/debug-logger';
 import {parseIEProcessArgs} from './util/args';
 import {andHandle} from './util/helpers';
 import {logger} from './util/logger';
 import {validateManifest} from './util/validations';
 
 import {STRINGS} from './config';
 
-const {DISCLAIMER_MESSAGE} = STRINGS;
+const {DISCLAIMER_MESSAGE, STARTING_IMPACT_FRAMEWORK, EXITING_IF} = STRINGS;
```
narekhovhannisyan, line 18: is exiting if, then starting if

jmcook1186:
hi @manushak I found a problem wit this PR - it doesn't output to `--stdout` unless `--debug` is raised.

src/util/debug-logger.ts
```diff
@@ -0,0 +1,73 @@
...
+  console.info = (...args: any[]) => debugLog(LogLevel.Info, args, debugMode);
+  console.warn = (...args: any[]) => debugLog(LogLevel.Warn, args, debugMode);
+  console.error = (...args: any[]) => debugLog(LogLevel.Error, args, debugMode);
+  console.debug = (...args: any[]) => debugLog(LogLevel.Debug, args, debugMode);
+};
+
+/**
+ * Sets the name of the currently executing plugin.
+ */
+const getExecutingPluginName = (pluginName: string) => {
```
narekhovhannisyan, line 31: seems it's setting the plugin name instead of getting

BTW it's better to incapsulate plugin variable, and these methods into single function

narekhovhannisyan (changes requested):
Please check comments above

manushak:
> Screenshot 2024-06-03 at 17 48 01
> this log appear whenever if is executed without adding --debug flag.

Can you provide the command that you ran? I don't get that message

jmcook1186:
@manushak  yea - I fixed the manifest issue first and then found the bug. Try with any working manifest.

jmcook1186 (approved):
sgtm after latest changes - thanks

src/index.ts
narekhovhannisyan: @manushak you got this wrong. in starting you have written impact framework, however exiting if. So for fixing this inconsistency please use if in both places.

manushak: Uh, I've got it now. The variable names are consistent with the message itself. 
@jmcook1186, maybe we need to rename 'Starting Impact framework' to 'Starting IF' to be consistent with 'Exiting IF'. What do you think?

manushak: @narekhovhannisyan  renamed

narekhovhannisyan (approved):

MariamKhalatova (changes requested):
@manushak 
`Syncing prameters` is shown at the beginning of the log and without DEBUG label. So no matter running with --debug or without it. The console.debug(SYNCING_PARAMETERS); was called at wrong place

Screenshot 2024-06-07 at 12 11 38 AM

MariamKhalatova:
@manushak we are getting ton of checking aggregation method logs. Each field should appear once.
`mock-obs-time-sync.yml` was used to reproduce.

Screenshot 2024-06-08 at 12 42 50 PM

Screenshot 2024-06-08 at 12 43 31 PM

MariamKhalatova:
@manushak when running if with debug flag, exhaust's logs are gone under `INFO: `.

Screenshot 2024-06-08 at 1 42 54 PM

Screenshot 2024-06-08 at 1 44 52 PM

manushak:
> @manushak we are getting ton of checking aggregation method logs. Each field should appear once. `mock-obs-time-sync.yml` was used to reproduce.
> 
> Screenshot 2024-06-08 at 12 42 50 PM Screenshot 2024-06-08 at 12 43 31 PM

@MariamKhalatova we get these huge logs because it traverses every input field.
@jmcook1186 Wdyt is this the right behavior?

MariamKhalatova:
@manushak It's aggregation log and it's called on each input (imagine what would happen with complex pipelines of 1000 nodes). There was memoize log function in code base, you can use that to log each one once.

jmcook1186:
@manushak @MariamKhalatova I think the aggregation check only needs to be displayed one time for each parameter (i.e. for each element in `inputs[0]`), as the operation is entirely idempotent - nothing can change between `inputs[i]` and `inputs[i+1]...inputs[N]` (or at least the probability is so low that it is not worth having the additional logs).

manushak:
@MariamKhalatova, I've fixed. Please test

src/lib/parameterize.ts
```diff
@@ -68,6 +70,25 @@ const Parameterize = () => {
     parametersStorage = parameters;
   };
 
+  /**
+   * Creates a handler for managing unique unit names.
+   */
+  const setUniqueUnitName = (() => {
```
narekhovhannisyan, line 76: we already have something like this in our codebase memoizedLog

you can reuse that with minor modifications to avoiding code duplication

jmcook1186:
what's the status of this @manushak @MariamKhalatova ? have @MariamKhalatova 's requested changes been addressed?

manushak:
> what's the status of this @manushak @MariamKhalatova ? have @MariamKhalatova 's requested changes been addressed?

the requested changes have been fixed, and are waiting for testing

MariamKhalatova (approved):

Issue #739, opened 2024-05-27
Cannot convert undefined or null to object


DimaRadionov:
I wanted to run a manifest file and received an error: Cannot convert undefined or null to object. 
## Description of the Error

I was developing a tool, which will track the co2 emmisions. When it all cames to run the manifest file, I received an error. 
## Expected Behaviour

Should run, cause I copied a simple manifest file 
## Actual Behaviour

error:     Cannot convert undefined or null to object
## Manifest File That Generated the Error

name: basic-demo
description:
tags:
initialize:
  plugins:
    teads-curve:
      path: '@grnsft/if-unofficial-plugins'
      method: TeadsCurve
      global-config:
        interpolation: spline
tree:
  children:
    child-0:
      defaults:
        cpu/thermal-design-power: 100
      pipeline:
        - teads-curve
      inputs:
        - timestamp: 2023-07-06T00:00
          duration: 1
          cpu/utilization: 20
        - timestamp: 2023-07-06T00:01
          duration: 1
          cpu/utilization: 80
        - timestamp: 2023-07-06T00:02
          duration: 1
          cpu/utilization: 20
## Runtime Info

[2024-05-27 09:05:51.402 AM] error:     Cannot convert undefined or null to object
TypeError:      Cannot convert undefined or null to object
    at Function.entries (<anonymous>)
    at listDependencies (C:\Users\e-doiv\AppData\Roaming\npm\node_modules\@grnsft\if\build\lib\environment.js:42:33)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async injectEnvironment (C:\Users\e-doiv\AppData\Roaming\npm\node_modules\@grnsft\if\build\lib\environment.js:49:26)
    at async impactEngine (C:\Users\e-doiv\AppData\Roaming\npm\node_modules\@grnsft\if\build\index.js:22:25)

jmcook1186:
Hi - sorry but I can't reproduce your error. You pasted the yaml flat, so I can't check whether the indentation was correct. This is how I ran your yaml data in my local IF and it executed fine.

```yaml
name: basic-demo
description:
tags:
initialize:
  plugins:
    teads-curve:
      path: '@grnsft/if-unofficial-plugins'
      method: TeadsCurve
      global-config:
        interpolation: spline
tree:
  children:
    child-0:
      defaults:
        cpu/thermal-design-power: 100
      pipeline:
        - teads-curve
      inputs:
[...]

PR #756, opened 2024-06-03
fix(src): check if packages empty so we don't return null

adamaucamp:
### Types of changes

Essentially just do a check if the returned packages from the environment with the `npm --list` command during setup actually returns anything, and return an empty object instead of null. 

see this issue for more (link)[https://github.com/Green-Software-Foundation/if/issues/739]

### DIFF
--- a/src/lib/environment.ts
+++ b/src/lib/environment.ts
@@ -53,7 +53,7 @@ const flattenDependencies = (dependencies: [string, PackageDependency][]) =>
 const listDependencies = async () => {
   const {stdout} = await execPromise('npm list --json');
   const npmListResponse: NpmListResponse = JSON.parse(stdout);
-  const dependencies = Object.entries(npmListResponse.dependencies);
+  const dependencies = Object.entries(npmListResponse.dependencies ?? {});
 
   return flattenDependencies(dependencies);
 };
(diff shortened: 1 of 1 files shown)


jmcook1186:
hi - thanks a lot for the PR!
pinging @narekhovhannisyan to review as he is closest to this aspect of the project.

narekhovhannisyan:
Hi @adamaucamp , thanks a lot for your contribution! 
With the current solution, we will create an empty array by calling object entries on the empty object. And then we have an unnecessary call of the `flattenDependencies` function.

So what do you think about solving the problem with the following strategy?

```ts
const listDependencies = async () => {
  const {stdout} = await execPromise('npm list --json');
  const npmListResponse: NpmListResponse = JSON.parse(stdout);

  if (npmListResponse.dependencies) {
    const dependencies = Object.entries(npmListResponse.dependencies);

    return flattenDependencies(dependencies);
  }

  return [];
};
```

adamaucamp:
Hi @narekhovhannisyan, that is a much better solution yes. I'll update the PR. thanks.

narekhovhannisyan (approved):

jmcook1186 (approved):

PR #790, opened 2024-06-07
Build `If-env`

manushak:
### Types of changes

- Buld if-env CLI to allow the user to setup a local environment to run a manifest file.
-


jmcook1186:
@manushak still reviewing this - seems good so far but an observation is that it takes quite a while to complete and during that time the user does not know what is happening - can we add some execution logs?

manushak:
> @manushak still reviewing this - seems good so far but an observation is that it takes quite a while to complete and during that time the user does not know what is happening - can we add some execution logs?

sure. I'll add logs

src/env/env.ts
```diff
@@ -0,0 +1,211 @@
...
+
+  await execPromise(`npm install ${packages.join(' ')}`, {
+    cwd: folderPath,
+  });
+}
+
+/**
+ * Updates package.json dependencies.
+ */
+async function updatePackageJsonDependencies(
```
narekhovhannisyan, line 171: why not arrow function?

src/env/env.ts
```diff
@@ -0,0 +1,211 @@
+#!/usr/bin/env node
+/* eslint-disable no-process-exit */
+import * as fs from 'fs/promises';
```
narekhovhannisyan, line 3: please move the file to src folder. We don't have specific folders for each script

src/env/env.ts
```diff
@@ -0,0 +1,211 @@
...
+const {
+  SUCCESS_MESSAGE,
+  FAILURE_MESSAGE,
+  FAILURE_MESSAGE_TEMPLATE,
+  FAILURE_MESSAGE_DEPENDENCIES,
+} = IF_ENV;
+
+const FOLDER_NAME = 'if-environment';
+
+type EnvironmentOptions = {
```
narekhovhannisyan, line 26: type can be moved

src/env/env.ts
```diff
@@ -0,0 +1,211 @@
...
+/**
+ * Gets depencecies with versions.
+ */
+const extractPathsWithVersion = (
+  plugins: any,
+  dependencies: string[]
+): {[path: string]: string} => {
+  const paths = Object.keys(plugins).map(plugin => plugins[plugin].path);
+  const uniquePaths = [...new Set(paths)].filter(path => path !== 'builtin');
+  const pathsWithVersion: {[path: string]: string} = {};
```
narekhovhannisyan, line 96: extract as a type

src/env/env.ts
```diff
@@ -0,0 +1,211 @@
...
+    await execPromise('npm init -y', {cwd: folderPath});
+  }
+
+  return packageJsonPath;
+}
+
+/**
+ * Installs packages from the specified dependencies in the specified folder.
+ */
+async function installDependencies(
```
narekhovhannisyan, line 155: this kind of helpers can be moved to utils

src/env/env.ts
manushak: in this case, where the if-env template manifest be placed? 
@jmcook1186 should it be moved to the` ./manifests/examples` folder with the name `if-env-template.yml`?

jmcook1186: no, it should just be saved to the top level `src` directory. `if-env` is not supposed to be run inside the `if` source repository

jmcook1186 (changes requested):
Thanks for this, it's nearly there, but there is also some behaviour we need to adjust:

1) When I run `if-env -m manifests/outputs/example-pipeline.yml` in the top-level `if` directory, I expect to see the dependencies listed in `example-pipeline.yml#execution` in the existing package.json in my IF folder. What I actually see is a new package.json added to /manifests/outputs/package.jon that *only* contains installed plugins as dependencies.

Think of this as `if-env` being a shortcut to running `npm add ...` for each plugin in the `initialize` block - the end result is a unified manifest that is ready to install and run the given manifest.

2) Similarly, when I run `if-env` with no params in a freshly created new folder, `if/temp` I expect to see a template manifest and package.json added to `if/temp`. What I actually see is the package.json added to `if/src` and an error message `faied to create the environment with the template manifest!`

This should work like `create-react-app` where in a fresh directory I can bootstrap my way to a dev environment with a package.json and template manifest so i can start building plugins.

manushak:
> Thanks for this, it's nearly there, but there is also some behaviour we need to adjust:
> 
> 1. When I run `if-env -m manifests/outputs/example-pipeline.yml` in the top-level `if` directory, I expect to see the dependencies listed in `example-pipeline.yml#execution` in the existing package.json in my IF folder. What I actually see is a new package.json added to `/manifests/outputs/package.json` that _only_ contains installed plugins as dependencies.
> 
> Think of this as `if-env` being a shortcut to running `npm add ...` for each plugin in the `initialize` block - the end result is a unified manifest that is ready to install and run the given manifest. So if i run `if-env -m manifests/outputs/example-pipeline.yml -install` I am ready to run `if-run -m example-pipeline.yml` without any further setup.
> 
> 2. Similarly, when I run `if-env` with no params in a freshly created new folder, `if/temp` I expect to see a template manifest and package.json added to `if/temp`. What I actually see is the package.json added to `if/src` and an error message `faied to create the environment with the template manifest!`
> 
> This should work like `create-react-app` where in a fresh directory I can bootstrap my way to a dev environment with a package.json and template manifest so i can start building plugins.
> 
> 3. Finally, get @narekhovhannisyan 's opinion on this too, but I think the template `package.json` should be the one we provide in the `if-plugin-template` repository (which will need to be updated to include `if-core`). Appreciate this isn't described in the original ticket.

@jmcook1186, the second point is problematic. The `process.cwd()` should return the working directory, but it doesn't; it returns the root directory of the project. It seems to be a common issue, but hopefully only locally. I've researched, and there isn't another way to get it. I've added `console.log(process.cwd())` in the `if-plugins` project in some of the plugins, and when running the manifest in the `IF`, the console logs the right directory. 
@narekhovhannisyan, have you found another way for this? You said you did something like this.

narekhovhannisyan:
@jmcook1186 @manushak I have checked the `if-env` briefly, and agree that having a freshly created package.json in `src` folder is not optimal. It starts to break the dependency management in the repository (nested package jsons are trying to override the existing one and vscode intellisense can't figure out dependencies either). 
For the second point haven't got much time to find solution for it, will continue my investigations.

narekhovhannisyan:
@jmcook1186 @manushak here is solution to find out from where script was called. Please apply changes needed to this snippet (execPromise).

```
const {exec} = require('node:child_process')

exec('pwd', console.log)
```

manushak:
> @jmcook1186 @manushak here is solution to find out from where script was called. Please apply changes needed to this snippet (execPromise).
> 
> ```
> const {exec} = require('node:child_process')
> 
> exec('pwd', console.log)
> ```

@narekhovhannisyan  it still returns the root folder of the project, aka `if`

narekhovhannisyan:
@jmcook1186 @manushak here are few feedbacks from me and @MariamKhalatova after brief testing

1. If manifest is updated and if-env is executed, the old package.json is not updated
2. "@grnsft/if": "0.4.0" should not be as a dependency in package.json file (our if module is not designed to work as a dependency.
3. In `initializeAndInstallLibs` function error is handled, however log is hardcoded and doesn't contain error message (generic non informative message)
4.  Use custom error instead
```
if (!dependencies.length) {
    throw new Error(FAILURE_MESSAGE_DEPENDENCIES);
  }
```

narekhovhannisyan:
@manushak I see the problem. It's because we are not calling directly file, instead we are using npm script to run which overrides the value.

narekhovhannisyan:
@manushak @jmcook1186 here is working solution:

1. in scripts folder create file pwd.sh with the content:
```bash
#!/bin/sh
CURRENT_PWD=$(pwd)
```

2. In `package.json` file change `if-env` script to following:
```json
"if-end": "bash ./pwd.sh && npx ts-node src/env.ts"
```

3. Access to CURRENT_PWD variable from src/env.ts file with `process.env.CURRENT_PWD`

narekhovhannisyan:
@manushak I have pushed the changes, please run locally to see if it's working on your machine

cc: @jmcook1186

jmcook1186:
Hi @narekhovhannisyan just pulled down the changes and tested locally, but I still see the same behaviour (i.e. run `if-env` from the `if` root and still see new `package.json` and dummy manifest added to `if/src`

narekhovhannisyan:
@jmcook1186 I have provided the key, not the whole solution. Manushak will continue building the remaining stuff.

jmcook1186 (approved):
Great work, @manushak - thanks

narekhovhannisyan:
@manushak @jmcook1186 can we move env-template to config ?

narekhovhannisyan:
@manushak let's keep env.ts file as clean as possible from functions. It should contain only the main function, all others can be sorted in utilities. And by doing that we will be able to write unit tests.

narekhovhannisyan (changes requested):
See comment above

narekhovhannisyan:
BTW please sort functions to appropriate utils, like npm related stuff to `utils/npm.ts`, fs related stuff to it's specific place and so on.

MariamKhalatova:
> 3\. In `initializeAndInstallLibs` function error is handled, however log is hardcoded and doesn't contain error message (generic non informative message)
> 4\. Use custom error instead
> 
> ```
> if (!dependencies.length) {
>     throw new Error(FAILURE_MESSAGE_DEPENDENCIES);
>   }
> ```

@manushak Is this fixed? Currently I don't see any changes related to this.

MariamKhalatova (changes requested):
@manushak Noticed such an issue. When running from outside of IF ,with prefix and cwd, unnecessary dependencies are added into the generated package.json file. 
Expecting to NOT see the framework dependencies there. Please check.

Used command: 
npm run --prefix=./if if-env -- -m manifests/testing/testing.yaml --cwd

Screenshot 2024-06-23 at 9 45 10 PM

package.json
```diff
@@ -85,4 +88,4 @@
   },
   "stability": "stable",
   "types": "src/index.d.ts"
-}
+}
```
narekhovhannisyan, line 91: missing newline

src/env.ts
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  console.log(SUCCESS_MESSAGE);
+  process.exit(0);
+};
+
+/**
+ * Gets the folder path of manifest file, dependencies from manifest file and install argument from the given arguments.
+ */
+const getOptionsFromArgs = async (commandArgs: {
```
narekhovhannisyan, line 60: is there any specific reason why these functions are here?

src/env.ts
```diff
@@ -0,0 +1,130 @@
...
+  manifest: string;
+  install: boolean | undefined;
+}) => {
+  const {manifest: manifestPath, install} = commandArgs;
```
narekhovhannisyan, line 64: to escape renaming, command args function can return manifestPath instead of manifest

narekhovhannisyan:
@manushak why in one case we are listing them and in another don't?

manushak:
@narekhovhannisyan, we decided that when running with the `--cwd` flag, it should add dependencies to the existing package.json. For example, if another manifest file already exists in the root directory, running the command should append dependencies so that both manifests can be run properly.

cc: @jmcook1186

package.json
jmcook1186: @narekhovhannisyan in cases like this, please just add the newline and commit. It's silly for these changes to involve >1 person.

jmcook1186:
```suggestion
}

```

jmcook1186:
```suggestion
}

```

jmcook1186:
Let's bring this to a close asap please. @manushak please address the outstanding comments as your priority task on IF. @narekhovhannisyan please approve unless there are changes that negatively affect the behaviour - no blocking on cosmetic issues that can be tidied up later.

manushak:
@jmcook1186, @narekhovhannisyan, I've pushed the latest changes. Some functions don't have unit tests yet, but I'll add them later. I couldn't add the custom Error from if-core for `throw new Error(FAILURE_MESSAGE_DEPENDENCIES);`. There is a problem with husky; I couldn't link `if-core` with `if`.

narekhovhannisyan:
@jmcook1186 all my feedback is about the architecture and consequential poor unit test coverage (huge amount of functions were in env.ts which is not covered with unit tests), nothing cosmetic there.

manushak:
> @manushak Noticed such an issue. When running from outside of IF ,with prefix and cwd, unnecessary dependencies are added into the generated package.json file. Expecting to NOT see the framework dependencies there. Please check.
> 
> Used command: npm run --prefix=./if if-env -- -m manifests/testing/testing.yaml --cwd
> 
> Screenshot 2024-06-23 at 9 45 10 PM

@MariamKhalatova, @narekhovhannisyan sorry, I got it wrong from the screenshot. I thought the selected dependencies were missing and you wanted them to persist. Thanks for noticing. It's fixed now.

package.json
narekhovhannisyan, line 91:
```suggestion
}

```

MariamKhalatova (approved):

PR #857, opened 2024-06-21
Build `If-check`

manushak:
### Types of changes

-  Add functionality to check if the provided manifest outputs are correct
- There are 2 cases to test:
- When you provide manifest file path: `npm run if-check — -m ./manifests/outputs/manifest.yaml`. If the outputs are the same as the generated manifest file, it should print `Files match!` or  
`Files do not match!
tree.children.child.outputs.0.cpu/energy
source: 0.08333333333333333
target: 0.1`
- When you provide a directory path where placed manifests files are: `npm run if-check — -d ./manifests/output/`. It should print every file name with the execution result


jmcook1186 (approved):

src/check.ts
```diff
@@ -0,0 +1,105 @@
...
+        const executedFile = file.replace(fileName, `re-${fileName}`);
+
+        logFailureMessage(error);
+        await removeFileFromDirectory(executedFile);
+      }
+    }
+  }
+};
+
+const executeCommands = async (manifest: string, cwd: boolean) => {
```
narekhovhannisyan, line 55: you know what to do

src/check.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  if (result.stdout) {
+    const logs = result.stdout.split('\n\n');
+    const successMessage = logs[logs.length - 1];
+
+    console.log(successMessage);
+  }
+};
+
+const logFailureMessage = (error: any) => {
```
narekhovhannisyan, line 92: this one too

src/check.ts
```diff
@@ -0,0 +1,105 @@
...
+  } else {
+    const directory = commandArgs.directory;
+    const files = await getYamlFiles(directory!);
+
+    for await (const file of files) {
+      const fileName = path.basename(file);
+      console.log(fileName);
+
+      try {
+        await executeCommands(file, true);
```
narekhovhannisyan, line 43: let;s try to escape having so many nested try catch blocks

src/util/fs.ts
```diff
@@ -11,3 +12,52 @@ export const isFileExists = async (filePath: string) => {
...
+    } else {
+      if (file.endsWith('.yml') || file.endsWith('.yaml')) {
+        yamlFiles.push(fullPath);
+      }
+    }
+  }
+
+  return yamlFiles;
+};
+
```
narekhovhannisyan, line 58: missing doc here

src/check.ts
manushak: > let;s try to escape having so many nested try catch blocks

This one is necessary because when we get an error from one manifest, it should not stop; it should continue checking the others.

src/util/fs.ts
```diff
@@ -11,3 +12,52 @@ export const isFileExists = async (filePath: string) => {
...
+ */
+export const removeFileFromDirectory = async (filePath: string) => {
+  await fs.unlink(filePath);
+};
+
+/**
+ * Gets all files that have either .yml or .yaml extension in the given directory.
+ */
+export const getYamlFiles = async (directory: string) => {
+  let yamlFiles: string[] = [];
```
narekhovhannisyan, line 39: have you tried like this/.

```
const files = await fs.readdir(directory);

  for (const file of files) {
    const fullPath = path.join(directory, file);
    const stat = await fs.lstat(fullPath);

    if (stat.isDirectory()) {
      return yamlFiles.concat(await getYamlFiles(fullPath));
    } 
    
    if (file.endsWith('.yml') || file.endsWith('.yaml')) {
      return yamlFiles.push(fullPath);
    }
  }

  return yamlFiles;
```

manushak: @narekhovhannisyan , I've tried this way, but I've got some errors and left my version.

manushak:
I'm closing this out because of a linter error that failed in this branch only and cannot be fixed. I will open a new PR instead.

PR #875, opened 2024-07-03
Make —stdout the default export option

manushak:
### Types of changes

- remove the `--stdout` flag
- add the `no-output` flag to prevent output to the console
- 2 cases to test 
    - `npm run if-run -- -m ./manifest-path`, it should output the result to the console
    - `npm run if-run -- -m ./manifest-path --no-output`, it should suppress output  to the console


src/config/config.ts
```diff
@@ -32,11 +32,11 @@ export const CONFIG = {
...
         description: '[path to a parameter file that overrides our defaults]',
       },
-      stdout: {
+      'no-output': {
         type: Boolean,
         optional: true,
-        alias: 's',
-        description: '[prints out to the console]',
+        alias: 'n',
+        description: '[privent output to the console]',
```
narekhovhannisyan, line 39:
```suggestion
        description: '[prevent output to the console]',
```

jmcook1186 (changes requested):
This doesn't seem to be working well. Running `if-run` without an output command *does* print the results to the console as expected, however, printing to the console now seems to be the only output option - saving to file is broken.

manushak:
> This doesn't seem to be working well. Running `if-run` without an output command _does_ print the results to the console as expected, however, printing to the console now seems to be the only output option - saving to file is broken.

@jmcook1186, if I understand correctly, now it is working as before (checked in the main branch). When the `outputs: ["yaml"]` is missing from the manifest, the `--output` flag does not work and it does not save. Do I need to change that behavior to save the executed manifest when the `--output` flag with a path is provided, even if `outputs: ["yaml"]` is missing?

jmcook1186:
in my local version adding `outputs: -yaml` in the manifest and using the `-o` command on the command line does not lead to a yaml file being saved, it still just prints o the console. Are you able to save files in the usual way?

manushak:
@jmcook1186 I run `npm run if-run -- -m ./manifests/plugins/divide/success.yml -o ./manifests/outputs/divide` and it works on my local (but beforehand I uncommented `outputs: ["yaml"]`)

jmcook1186:
ok, @manushak it's working for me now too - i must have had something wrong in my setup. Sorry, and thanks!

narekhovhannisyan (approved):

PR #913, opened 2024-07-16
Add `explain` feature

manushak:
### Types of changes

- add `explain` and `explainer` properties in the manifest
- to test add `explainer: true` in the manifest file and execute it, the necessary data will be provided in the execute section.

### DIFF
--- a/src/__tests__/if-run/lib/explain.test.ts
+++ b/src/__tests__/if-run/lib/explain.test.ts
@@ -1,0 +1,86 @@
+import {explain, addExplainData} from '../../../if-run/lib/explain';
+
+describe('lib/explain: ', () => {
+  it('successfully adds explain data if `inputs` and `outputs` of `metadata` are `undefined`.', () => {
+    const mockData = {
+      pluginName: 'divide',
+      metadata: {kind: 'execute', inputs: undefined, outputs: undefined},
+      pluginData: {
+        path: 'builtin',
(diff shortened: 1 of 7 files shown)


narekhovhannisyan:
Screenshot 2024-07-19 at 17 15 44
@manushak It will be better to have some default messages for the cases where information is not available.

jmcook1186:
Agree with @narekhovhannisyan - maybe we can use:
`x not available - please set x in the initialize block for y plugin`

manushak:
> Agree with @narekhovhannisyan - maybe we can use: `x not available - please set x in the initialize block for y plugin` @manushak

@jmcook1186 here `undefined` means that the user hasn't set `inputs` and `outputs` in the `parameter-metadata`.
Since it is not possible to add inputs and outputs to the manifest if they have undefined values (as undefined properties are skipped from the manifest), I've added the string undefined. I see two solutions for this: either add empty [] or skip those properties from the manifest if they have undefined values. What do you think?

jmcook1186:
In that case, maybe it's better just to leave them as `undefined` and just make it clear in the documentation that the remedy is to add parameter metadata in the `initialize` block.

jmcook1186 (approved):
lgtm, pending update to aggregation method as discussed in #916

narekhovhannisyan (approved):

Issue #1111, opened 2025-01-10
Add energy-consumption-calculator plugin to builtins


kaito-kameda:
**What**
A builtin plugin that calculates energy consumption from wattage and duration. 

**Why**
The calculation of SCI score requires energy consumption. However, in many cases, the data obtained from the machine in use is wattage, so it is necessary to calculate energy consumption based on wattage and duration. This is a common calculation, but currently it is calculated using two plugins, 'multiply' and 'divide', which is a bit cumbersome. Creating a plugin that directly calculates energy consumption from wattage and duration would allow the manifest to be written more concisely.

**Context**

**Prerequisites/resources**
None

**SoW (scope of work)**
- [ ] new directory in builtins with implementation
- [ ] test cases added
- [ ] README in new directory documenting plugin usage

**Acceptance criteria**
- [ ] plugin calculates energy consumption from given wattage and duration

**Given** the plugin works as intended
**When** I run the following manifest

```yaml
name: energy-consumption-calculater demo
description: successful path
tags:
initialize:
  plugins:
    calculateEnergyConsumption:
      method: CalculateEnergyConsumption
      path: "builtin"
      config:
        input-parameter: "wattage"
        output-parameter: "energy-consumption"
tree:
  children:
    child:
      pipeline:
        compute:
          - calculateEnergyConsumption
      inputs:
        - timestamp: 2023-08-06T00:00
          duration: 3600
          wattage: 100
```

**Then** I get the following output:

```yaml
name: energy-consumption-calculater demo
description: successful path
tags:
initialize:
  plugins:
    calculateEnergyConsumption:
      method: CalculateEnergyConsumption
      path: "builtin"
      config:
        input-parameter: "wattage"
        output-parameter: "energy-consumption"
tree:
  children:
    child:
      pipeline:
        compute:
          - calculateEnergyConsumption
      inputs:
        - timestamp: 2023-08-06T00:00
          duration: 3600
          wattage: 100
          energy-consumption: 0.1
```

jmcook1186:
Hi @kaito-kameda - responded on the linked PR about this. Will leave open for a day or so for you to respond if you wish, but I think this is better maintained in your own repository and shared via the explorer.

PR #1112, opened 2025-01-10
Add energy-consumption-calculator plugin

kaito-kameda:
### Types of changes

- 
- 

#1111

### DIFF
--- a/manifests/examples/builtins/energy-consumption-calculator/success.yml
+++ b/manifests/examples/builtins/energy-consumption-calculator/success.yml
@@ -1,0 +1,21 @@
+name: energy-consumption-calculater demo
+description: a manifest demonstrating how to use the energy-consumption-calculater feature
+tags:
+initialize:
+  plugins:
+    calculateEnergyConsumption:
+      method: CalculateEnergyConsumption
+      path: 'builtin'
+      config:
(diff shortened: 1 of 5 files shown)


kaito-kameda:
Closes #1111

jmcook1186:
Hi @kaito-kameda - thanks a lot for your work on this - indeed this is a commonly used pipeline of operations - abstracting it into its own plugin is a convenient way to make manifests less verbose. 

However, I don't think this belongs in the IF builtins - it's better maintained in your own repository and shared via the explorer (https://explorer.if.greensoftware.foundation).

The reason why I think this is that we aim to keep the IF standard library minimal and as low level as possible. Combinations of builtins and higher level abstractions can then be shared either as community plugins or even just as reusable pipelines (i.e. yaml that can be pasted into manifests). This approach balances minimizing the complexity and maintenance burden for IF itself with encouraging a decentralized community of builders to share things independently of the IF core team.

If you want to share the plugin, see here https://wiki.greensoftware.foundation/how-to-add-plugins.

I'd also suggest using a more specific name for the plugin - `energy-consumption-calculator` could mean a lot of things - people might expect a full end to end calculation from observation -> consumption. Something like "power-to-energy" is less ambiguous.

PR #1126, opened 2025-03-09
fix(src): fix osInfo

kariya-mitsuru:
### Types of changes

In `getLinuxInfo`, `lsb_release -a` was used, but since the separator in this command's output is a tab character rather than a space, the output was not retrieved correctly.
Additionally, since `getLinuxInfo` only needed Distributor ID and Description, the option was changed from -a to -id.

cf. https://refspecs.linuxfoundation.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/lsbrelease.html

In `getWindowsInfo`, `systeminfo` was used, but its output is affected by the current code page.  As a result, in non-English environments, the output could not be retrieved correctly, leading to errors.  This was fixed by using PowerShell's `Get-WmiObject`, ensuring correct retrieval regardless of the current code page.  Additionally, an extra CR (\r) at the end of the name was removed.

Furthermore, `getLinuxInfo`, `getWindowsInfo`, and `getMacVersion` all used `execPromise`, but since a shell is not necessary to execute these commands, a new `execFilePromise` was created and used instead.

### DIFF
--- a/src/__tests__/if-run/util/os-checker.test.ts
+++ b/src/__tests__/if-run/util/os-checker.test.ts
@@ -10,7 +10,7 @@ jest.mock('os', () => ({
   release: () => 'm.m.m',
 }));
 jest.mock('../../../common/util/helpers', () => ({
-  execPromise: async () => {
+  execFilePromise: async () => {
     if (process.env.KIND === 'darwin' && process.env.REJECT === 'true')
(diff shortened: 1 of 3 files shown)


jmcook1186:
sgtm! thanks @kariya-mitsuru

PR #1134, opened 2025-05-05
fix(src): fix getProcessStartingTimestamp function

kariya-mitsuru:
### Types of changes

The `getProcessStartingTimestamp` function was using hrtime but wasn't correctly calculating the process start time.
Therefore, modify it to use uptime for the calculation instead.

Additionally, change it to use the standard `Date` object instead of luxon, as the rich features provided by luxon weren't necessary.

Note: When `hrtime` is called without arguments, it returns the elapsed time from some point (not specified in the API), rather than returning the value of the process start time.

### DIFF
--- a/src/if-run/lib/environment.ts
+++ b/src/if-run/lib/environment.ts
@@ -1,5 +1,3 @@
-import {DateTime} from 'luxon';
-
 import {osInfo} from '../util/os-checker';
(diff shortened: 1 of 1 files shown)


jmcook1186 (approved):
seems ok to me - thanks!

PR #1140, opened 2025-05-12
fix(config): refine test configuration

kariya-mitsuru:
### Types of changes

The `modulePathIgnorePatterns` in `jest.config.js` was set to `./build`, which caused all modules to be ignored and tests not to be executed when the git working directory was under `build` (as it was in my environment). Therefore, I changed it to `<rootDir>/build/`.
(Since `modulePathIgnorePatterns` is a regex and is checked against the absolute path of the target, `./build` matches "paths that are under a directory with any name containing one or more characters (i.e., any directory other than the root directory) and start with `build`.)

Also made similar changes to `coveragePathIgnorePatterns` in `jest.config.js`.
(I don't think the git working directory will match `src/.*/config/.*` or `src/.*/types/.*` though...)

Additionally, both the `test` and `coverage` scripts in `package.json` had the `--testPathPattern=src/__tests__/` command line option specified. So, I added the pattern `<rootDir>/src/__tests__/**/*.[jt]s` (which is a Glob pattern, not a regex) to `testMatch` in `jest.config.js` and removed the command line option.
(For the same reason mentioned above, I included `<rootDir>` at the beginning.)
This enables running tests with a narrowed scope during test execution, such as `npm test -- --testPathPattern=if-run`.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -2,6 +2,7 @@
 module.exports = {
   preset: 'ts-jest',
   testEnvironment: 'node',
+  testMatch: ['<rootDir>/src/__tests__/**/*.[jt]s'],
   transform: {
     '^.+\\.ts?$': [
       'ts-jest',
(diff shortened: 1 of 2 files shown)


jmcook1186 (approved):
seems ok to me!

